PR #14, opened 2022-04-20
Introduce ADRs and add one on node and client IDs

matejpavlovic:

### DIFF
--- a/docs/architecture-decision-records/adr-001-node-and-client-id-types.md
+++ b/docs/architecture-decision-records/adr-001-node-and-client-id-types.md
@@ -1,0 +1,43 @@
+# ADR 1: Types of Node IDs and Client IDs
+
+## Context
+
+Each node and each client in Mir is uniquely identified by an ID.
+The types of the IDs are currently defined in [/pkg/types/types.go](/pkg/types/types.go)
+as unsigned 64-bit integers.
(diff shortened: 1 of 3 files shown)


dnkolegov (approved):

docs/architecture-decision-records/adr-001-node-and-client-id-types.md
```diff
@@ -0,0 +1,43 @@
...
+like, for example, the port number the node is listening on.
+
+Such numeric IDs, however, create a hurdle when using Mir in systems where nodes' and clients' identifiers
+are already fixed (e.g. if Mir is only used to implement one component of a bigger system
+like [Eudico](https://github.com/filecoin-project/eudico))
+and cannot easily be converted to and from this type (e.g. public keys).
+Another translation layer between system-native IDs and IDs used inside Mir would be necessary.
+Using strings instead enables a more direct use of the system's native IDs,
+as almost anything can be converted to a string.
+When it comes to the (very frequent) use of IDs as map keys, strings can still be used for this purpose.
```
sergefdrv, line 19: I see at least the following advantages of using numerical IDs for identifying replicas:
 - it is conventional in permissioned BFT consensus protocols, e.g. in PBFT the primary replica's ID is derived from the view number
 - it can be more efficiently represented on wire, e.g. in VarInt encoding

PR #15, opened 2022-04-21
Accept ADR 1: String types for node and client IDs

matejpavlovic:

### DIFF
--- a/docs/architecture-decision-records/README.md
+++ b/docs/architecture-decision-records/README.md
@@ -8,4 +8,4 @@ When an ADR is added or updated, the following table of contents must manually b
 
 ## Table of contents
 
-- [001: Node and client ID types](adr-001-node-and-client-id-types.md) (proposed)
+- [001: Node and client ID types](adr-001-node-and-client-id-types.md) (Accepted Apr 21, 2022)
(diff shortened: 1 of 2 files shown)


dnkolegov (approved):

matejpavlovic:
> it is conventional in permissioned BFT consensus protocols, e.g. in PBFT the primary replica's ID is derived from the view number

That is true, but the id is abstracted away (by its dedicated type) for the protocol implementation. I'd strongly discourage the protocol to rely on any semantics of the IDs and treat them as opaque values that it only compares to each other (and maybe even for that we should introduce a dedicated method to avoid problems with potential deep / shallow comparisons). If the protocol needs to attach meaning to the IDs, it should manage it itself. For PBFT, this is very easy, since the membership is an ordered list and it suffices to use the position in that list.

> it can be more efficiently represented on wire, e.g. in VarInt encoding

That is also true and a valid point. I do not think however that it makes a significant performance difference and the simplicity of not having to add a layer of ID translation at Mir level is well worth the few extra bytes. If we want to squeeze out the last bit of performance out of it, it won't be hard to switch to integers.

sergefdrv (approved):
I don't have particularly strong arguments against the proposal for now, so I approve it.

Some further observations:
 - The wallet address alone may not be enough to sufficiently identify a remote node across Mir framework for all purposes, e.g. due to possible TLS certificate or consensus public key rotation, network address changes, etc. Therefore we'd probably need to combine it with a configuration number: `(configNr, nodeIdentity)` where `nodeIdentity` represents the wallet address as a string. 
 - Numerical IDs have some properties that string IDs don't have, e.g. one can do arithmetic with numerical IDs like comparing with `N`, `viewNr mod N`, etc.
 - `(configNr, nodeNr)` just looks to me more natural than `(configNr, nodeIdentity)`

PR #18, opened 2022-04-21
Use context instead of classic channel

### LINKED ISSUE #6: Use a Context instead of an "exit channel" in Node.Run()
Make the `Node.Run()` function
https://github.com/filecoin-project/mir/blob/946689a112983e6384d6c697c01de00d0f433d14/node.go#L176
accept a Context instead of an "exit channel". A Context fulfils the same purpose, but is the idiomatic way of achieving it in Go.

dnkolegov:
This PR fixes #6.


matejpavlovic (approved):
Looks good, thanks!

Issue #9, opened 2022-04-20
Basic ISS checkpoints


matejpavlovic:
Create certified checkpoints and produce debug output when they are created.

matejpavlovic:
Nice job!

PR #34, opened 2022-04-27
Implement collecting checkpoint certificates

sergefdrv:
This pull request implements generating and validating Checkpoint message signatures, as well as collecting checkpoint certificates.

Relates to #9


pkg/iss/checkpoint.go
```diff
@@ -179,11 +213,17 @@ func (ct *checkpointTracker) stable() bool {
 }
 
 func (ct *checkpointTracker) announceStable() *events.EventList {
```
matejpavlovic, line 215:
```go
// Assemble a multisig certificate from the received signatures.
```

matejpavlovic (approved):
Very nice, thanks!

PR #38, opened 2022-04-29
Fix hanging transport layer shutdown in tests

### LINKED ISSUE #36: Testing sometimes hangs
Sometimes `make test` hangs, e.g. [here](https://github.com/filecoin-project/mir/runs/6215449350).

matejpavlovic:
The transport layer is writing messages that it receives over the
network to a channel, from which the node implementation reads them.
When the node is shutting down, it stops reading incoming messages from
this channel. However, if there are still incoming messages on the wire
at that time, the transport layer tries to write them to the channel and
blocks (since nobody reads any more). This prevented a clean shutdown of
the transport layer.

Now the connecting node explicitly cancels the context of the connection
when disconnecting and the node being connected uses a select statement,
waiting for this event alongside trying to write the message to the
processing channel.

Fixes #36

### DIFF
--- a/pkg/deploytest/deployment.go
+++ b/pkg/deploytest/deployment.go
@@ -115,7 +115,11 @@ func NewDeployment(testConfig *TestConfig) (*Deployment, error) {
 		case "fake":
 			transport = fakeTransport.Link(t.NewNodeIDFromInt(i))
 		case "grpc":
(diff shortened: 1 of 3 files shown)


sergefdrv (approved):
:+1: Thanks!

PR #44, opened 2022-05-03
Batch signature verification

matejpavlovic:
Multiple signatures can now be verified using a single event.
This is useful for verifying, for example, multisig certificates.


sergefdrv (commented):
`VerifyNodeSig` is essentially the same as `VerifyNodeSigBatch`, but with only a single signature in the batch. Why don't we just use the latter and avoid code duplication?

matejpavlovic:
Yes I was thinking about the same, just wanted to avoid the necessary refactoring for now, but you're right, the earlier the better. I'll push an update.

sergefdrv (approved):

PR #55, opened 2022-05-09
ISS: Fix delivery order bug

matejpavlovic:


sergefdrv:
@matejpavlovic Do I understand correctly that follow-up events are not guaranteed to be processed immediately next after the event they are attached to?

matejpavlovic:
Yes that's true. They are only guaranteed to be processed after the event they are attached to, but not immediately after. If processing "immediately after" (for events directed to the same module) turns out to be necessary, it should not be hard to modify the Mir framework to provide this guarantee.

PR #61, opened 2022-05-12
ISS: Fix quorum sizes for any n > 3f

sergefdrv:
The total number of replicas does not have to be exactly 3f+1 for some
natural number f. Fix the helper functions to calculate strong and
weak Byzantine quorums such that they give correct result in any case.


matejpavlovic (approved):
👍

PR #65, opened 2022-05-18
Use Context instead of ExitC in Node workers

matejpavlovic:
This approach is more idiomatic and thus easier to read and understand.


dnkolegov (approved):

sergefdrv (approved):
:+1:

PR #66, opened 2022-05-19
Fix transitioning into new epoch

### LINKED ISSUE #2: Improve epochFinished predicate in ISS
Resolve this TODO:
https://github.com/filecoin-project/mir/blob/946689a112983e6384d6c697c01de00d0f433d14/pkg/iss/iss.go#L666
The changes in the code should be rather straight-forward, but addressing this issue is a good opportunity to get to know the code.

sergefdrv:
This pull request fixes a bug introduced in c01d2e8 which broke `ISS.epochFinished` and resolves #2. It also adjusts the default ISS configuration such that there is at least one integration test that has to order requests across multiple epochs.

### DIFF
--- a/pkg/iss/config.go
+++ b/pkg/iss/config.go
@@ -8,6 +8,7 @@ package iss
 
 import (
 	"fmt"
+
 	t "github.com/filecoin-project/mir/pkg/types"
 )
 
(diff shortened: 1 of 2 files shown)


matejpavlovic (approved):
👍

PR #67, opened 2022-05-20
Timer module

### LINKED ISSUE #7: Timer module
Design and implement a Mir-native time module that would replace the explicit handling of `Tick` events by the protocol.

matejpavlovic:
Implements [ADR 2](https://github.com/filecoin-project/mir/blob/main/docs/architecture-decision-records/0002-timer-module.md) and resolves #7 .


pkg/timer/timer.go
```diff
@@ -0,0 +1,93 @@
...
+
+			// Try to write events to notifyChan until the context is canceled.
+			select {
+			case notifyChan <- events:
+			case <-ctx.Done():
+			}
+
+			// Wait for the defined period of time (or until the context is canceled)
+			select {
+			case <-time.After(time.Duration(period)):
```
sergefdrv, line 69: I think it is better to use a channel returned from [`time.Tick`](https://pkg.go.dev/time#Tick), because otherwise the interval may not be approximately equal to `period`.

pkg/modules/timer.go
```diff
@@ -0,0 +1,38 @@
...
+	// Delay writes events to notifyChan after delay.
+	// If context is canceled, no events might be written.
+	Delay(ctx context.Context, events *events.EventList, delay t.TimeDuration, notifyChan chan<- *events.EventList)
+
+	// Repeat periodically writes events to notifyChan with the given period.
+	// The repeated producing of events is associated with the retention index retIndex.
+	// The first write of events to notifyChan happens immediately after invocation of Repeat.
+	// Repeat stops writing events to notifyChan when
+	// 1) ctx is canceled or
+	// 2) GarbageCollect is called with an argument greater than the associated retIndex.
```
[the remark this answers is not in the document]
matejpavlovic: Good point, you're right. Will update.

pkg/modules/timer.go
```diff
@@ -0,0 +1,38 @@
...
+		retIndex t.TimerRetIndex,
+		notifyChan chan<- *events.EventList,
+	)
+
+	// GarbageCollect cancels the effect of outdated calls to Repeat.
+	// When GarbageCollect is called, the Timer stops repeatedly producing events associated with a retention index
+	// smaller than retIndex.
+	// Values of retIndex over multiple calls to GarbageCollect must be non-decreasing.
+	// If GarbageCollect is called with a retIndex smaller that a value used in a previous call,
+	// the behavior of the Timer is undefined.
```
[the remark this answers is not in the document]
matejpavlovic: Good point, you're right. Will update.

pkg/iss/pbftconfig.go
```diff
@@ -10,12 +10,12 @@ type PBFTConfig struct {
...
 
-	// The maximum number of logical time ticks between two proposals of new batches during normal operation.
+	// The maximum time duration between two proposals of new batches during normal operation.
 	// This parameter caps the waiting time in order to bound latency.
-	// When MaxProposeDelay ticks have elapsed since the last proposal,
+	// When MaxProposeDelay has elapsed since the last proposal,
 	// the protocol tries to propose a new request batch, even if the batch is not full (or even completely empty).
 	// Must not be negative.
-	MaxProposeDelay int
+	MaxProposeDelay t.TimeDuration
```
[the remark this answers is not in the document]
matejpavlovic: Same as above.

pkg/timer/timer.go
sergefdrv: This approach will send events with interval strictly grater than `duration` because `time.After` cannot close the returned channel before the passed duration elapses (in practice it will always take a little longer) and the other select statement in the loop may also take some time. 

> Tick ... [leaks memory](https://pkg.go.dev/time#Tick)

You're right, let's use `time.Ticker` instead.

sergefdrv: If we use `time.Ticker` then we'll have events sent with interval `period` on average.

pkg/iss/config.go
```diff
@@ -47,13 +48,13 @@ type Config struct {
...
-	// The maximum number of logical time ticks between two proposals of an orderer, where applicable.
+	// The maximum time duration between two proposals of an orderer, where applicable.
 	// For orderers that wait for a request batch to fill before proposing it (e.g. PBFT),
 	// this parameter caps the waiting time in order to bound latency.
-	// When MaxProposeDelay ticks have elapsed since the last proposal made by an orderer,
+	// When MaxProposeDelay has elapsed since the last proposal made by an orderer,
 	// the orderer proposes a new request batch, even if the batch is not full (or even completely empty).
 	// Must not be negative.
-	MaxProposeDelay int
+	MaxProposeDelay t.TimeDuration
```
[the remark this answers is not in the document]
sergefdrv: I think it's more convenient for the external code to have conventional `time.Duration` in the configuration structures. We can use the custom type internally.

pkg/timer/timer.go
```diff
@@ -0,0 +1,98 @@
...
+			select {
+			case <-ticker.C:
+			case <-ctx.Done():
+				// If the context is canceled before the delay elapses, stop the repetition.
+				break
+			}
+		}
+
+		// Stop the ticker so it can be garbage-collected.
+		ticker.Stop()
```
sergefdrv, line 83: It would be nicer to arrange this call with a `defer` statement right after creating the ticker

sergefdrv (approved):
:+1:

pkg/timer/timer.go
matejpavlovic: 👍

sergefdrv (approved):

PR #68, opened 2022-05-20
State transfer

### LINKED ISSUE #11: State transfer to a delayed node
After a node is delayed, it needs to be able to catch up and fetch the necessary protocol and app state from other nodes.

sergefdrv:
This pull request resolves #11. Tested on  chat demo with the following patch:

```diff
diff --git a/pkg/iss/iss.go b/pkg/iss/iss.go
index 68d38ed..ae67d31 100644
--- a/pkg/iss/iss.go
+++ b/pkg/iss/iss.go
@@ -665,6 +665,9 @@ func (iss *ISS) applyMessageReceived(messageReceived *eventpb.MessageReceived) *
 	case *isspb.ISSMessage_StableCheckpoint:
 		return iss.applyStableCheckpointMessage(msg.StableCheckpoint, from)
 	case *isspb.ISSMessage_Sb:
+		if iss.ownID == t.NodeID("1") {
+			return &events.EventList{}
+		}
 		return iss.applySBMessage(msg.Sb, from)
 	case *isspb.ISSMessage_RetransmitRequests:
 		return iss.applyRetransmitRequestsMessage(msg.RetransmitRequests, from)
```

Closes https://github.com/protocol/consensuslab/issues/69
Closes https://github.com/protocol/consensuslab/issues/62


pkg/iss/iss.go
```diff
@@ -609,8 +609,34 @@ func (iss *ISS) applyStableCheckpoint(stableCheckpoint *isspb.StableCheckpoint)
...
+		// Prune old entries from WAL. The entries to prune
+		// are determined according to the retention index
+		// which is derived from the epoch number the new
+		// stable checkpoint is associated with.
 		eventsOut.PushBack(events.WALTruncate(t.WALRetIndex(stableCheckpoint.Epoch)))
 
+		// Clean up the global ISS state from all the epoch
+		// instances that are associates with epoch numbers
+		// less than the epoch number the new stable
+		// checkpoint associated with.
```
matejpavlovic, line 639:
```suggestion
		// lower than the epoch number of the new stable checkpoint.
```

protos/eventpb/eventpb.proto
```diff
@@ -177,6 +178,10 @@ message AppSnapshot {
   bytes  data  = 2;
 }
 
+message AppRestoreState {
+  bytes  data  = 2;
```
matejpavlovic, line 182: Is there a reason to start counting at 2 or is it a typo?

pkg/iss/iss.go
```diff
@@ -703,6 +705,57 @@ func (iss *ISS) applyCheckpointMessage(message *isspb.Checkpoint, source t.NodeI
...
+
+	// Attach a follow-up event that activates SB instances of the
+	// new epoch after the application module has restored the
+	// state from the snapshot.
+	restoreEvent.FollowUps(iss.initOrderers().Slice())
+
+	// Attach a follow-up event that applies any message buffered
+	// for the new epoch after the application module has restored
+	// the state from the snapshot.
+	restoreEvent.FollowUps(iss.applyBufferedMessages().Slice())
```
matejpavlovic, line 754: Do these need to be follow-up events?
It seems to me that maybe the events could just be returned in a list.
Events in a single list that are destined to the same module are guaranteed to be applied in FIFO order to that module.

matejpavlovic (approved):
Very nice! Found just a few typos and have one question.

pkg/iss/iss.go
sergefdrv: You seem to be right, the new SB instances would't be able to interact with the application module anyway until the state restore event is fully handled, so there's apparently no need to delay activation of the new epoch instance.

sergefdrv: However, what if the new epoch instance happens to deliver a new batch (e.g. while processing the buffered events) before the restore state event reaches the application module's event queue?

sergefdrv:
Summary of [changes](https://github.com/filecoin-project/mir/compare/8a8f022d7a2b837b677a1d36c217a3566ce2c7c8..36821f2d13121512c7f72ddecbc3ec4c5eee1d3d): fixed typos

pkg/iss/iss.go
matejpavlovic: That doesn't seem possible to me, since it would have to happen as a direct result of the application of the `StableCheckpoint` message and precede the `AppRestoreState` event in the returned event list. By the time any other event is processed by the protocol, all the events resulting from the application of `StableCheckpoint` are already queued up.

sergefdrv: > By the time any other event is processed by the protocol, all the events resulting from the application of StableCheckpoint are already queued up.

Is it guaranteed that all events returned by a module always get queued to the destination event queues in order, even those that are returned as a result of processing by that module subsequent incoming events?

sergefdrv: This seems to be the case as long as `Node.workChans.workItemInput` channel is not buffered.

sergefdrv:
Summary of [change](https://github.com/filecoin-project/mir/compare/36821f2d13121512c7f72ddecbc3ec4c5eee1d3d..d868e2541688e7049f6e5fd647470d804f63bd72): removed deferring of epoch instance activation upon restoring state snapshot

pkg/iss/iss.go
matejpavlovic: I think it would even work if the channel was buffered. Outgoing events from a single module (in this case the `Protocol` module) are never reordered with respect to each other. A single thread injects the events in the `Protocol` module and the very same thread collects the resulting events. Thus, any result of processing the `StableCheckpoint` message (technically the `MessageReceived` event containing the `StableCheckpoint`) will be written to `workItemInput` (buffered or not) before any result of processing subsequent events, as they are handled by the same thread as the `StableCheckpoint`.

matejpavlovic: Note that this is not necessarily true for follow-up events. A follow-up event is only guaranteed to be processed after the event it is attached to, but not necessarily immediately after it. Arbitrary events, even originating in the same module, may pass in between.

PR #69, opened 2022-05-24
Use Timer for view change

matejpavlovic:


sergefdrv (approved):
Nicely structured PR!

PR #82, opened 2022-05-28
Module generalization

matejpavlovic:

### DIFF
--- a/docs/architecture-decision-records/0003-generalize-modules.md
+++ b/docs/architecture-decision-records/0003-generalize-modules.md
@@ -24,14 +24,14 @@ This approach:
 
 ## Considered Options
 
-* **Active and passive modules.**
+* **Active and passive module types.**
   Generalize the concept of modules, such that:
   - Modules have a unified interface for processing events,
     where each module internally inspects the incoming events and decides on the appropriate action.
     The `Protocol` and some other modules already do have such a unified interface - use it for other modules as well.
   - Instead of dedicated slots for each type of module, use a variable number of generic modules of only 2 types:
     - `PassiveModule`:
-      A module that simply transforms input events to output events whenever invoked by Mir.
+      A module that simply defines the logic for transforming input events to output events whenever invoked by Mir.
       Never injects external events in the system.
       A passive module has the following interface (corresponding to some existing modules, like `Protocol`):
       ```go
@@ -47,11 +47,9 @@ This approach:
       }
       ```
     - `ActiveModule`:
-      Like the `PassiveModule`, the `ActiveModule` consumes input events. However,
-      unlike the `PassiveModule`, the `ActiveModule` does not synchronously return receive input
-      and return output in a function call.
-      Instead, the `ActiveModule` asynchronously reads input events from a channel
+      The `ActiveModule` asynchronously reads input events from a channel
       and writes produced events to an output channel.
+      
       In an `ActiveModule`, the produced events might or might not depend on the input events.
       This makes the `ActiveModule` suitable for injecting external events
       (e.g., reception of a message over the network) in the Node.
--- a/docs/architecture-decision-records/0003-generalize-modules.md
+++ b/docs/architecture-decision-records/0003-generalize-modules.md
@@ -1,0 +1,112 @@
+# Generalize Modules
+
+* Status: proposed
+* Deciders: @matejpavlovic, @sergefdrv, @dnkolegov
+* Date: 2022-05-28 (last update)
+
+## Context and Problem Statement
+
+The current architecture defines several kinds of modules, each with a module-specific interface.
+There is also exactly one of each kind of module
+(i.e., one `Protocol` module instance, one `App`, one `ClientTracker`, etc.).
+The event handling code then takes each module and wraps it almost the same simple interface
+(consuming input events and producing output events) using a module-specific wrapper.
+This wrapper almost always ends up being just a simple de-multiplexer that,
+based on the incoming event type, calls the appropriate module-specific method of the module.
+
+This approach:
+- Pollutes the Node's event handling logic by the above-mentioned wrappers (which can be considered boilerplate code).
+  Whenever a new type of module is being defined, the core Mir code needs to be modified
+  to add the corresponding wrapper.
+- Unnecessarily constrains the user of Mir to use exactly one of each type of module, 
+  even if a protocol might naturally be split up in several modules working in parallel.
+- The selection of the types of modules supported is tailored to a SMR application, but not necessarily to other uses.
+
+## Considered Options
+
+* **Active and passive modules.**
+  Generalize the concept of modules, such that:
+  - Modules have a unified interface for processing events,
+    where each module internally inspects the incoming events and decides on the appropriate action.
+    The `Protocol` and some other modules already do have such a unified interface - use it for other modules as well.
+  - Instead of dedicated slots for each type of module, use a variable number of generic modules of only 2 types:
+    - `PassiveModule`:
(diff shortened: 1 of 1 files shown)


docs/architecture-decision-records/0003-generalize-modules.md
```diff
@@ -0,0 +1,112 @@
...
+        // the module will only produce a bounded number of output events and only in reaction to processing input events.
+        // In other words, if BackOff has been called, ctx has not been canceled, and the input channel is not written to,
+        // the module will eventually stop producing output events.
+        //
+        // This function is necessary for guaranteeing that the internal Node event buffers do not grow indefinitely.
+        // This could potentially happen when external events are injected into the system through active modules
+        // faster than the system can process them.
+        // In such a case, the Node asks all active modules to back off
+        // until enough events in the internal buffers have been processed.
+        BackOff(ctx context.Context)
```
sergefdrv, line 83: Assuming `Run` spawns a new goroutine to receive and produce events, a concurrent invocation to `BackOff` alone cannot strictly ensure a predictable boundary on the number of events produced until the invocation takes effect. To put an active module on hold, I suggest to simply stop receiving from `eventsOut` instead.

sergefdrv (changes requested):

docs/architecture-decision-records/0003-generalize-modules.md
sergefdrv: > Technically, `Run` need not itself spawn a new goroutine, but I'd expect it to be called as a separate goroutine. But that's almost the same thing for many practical purposes, including this one.

It would be better to clarify in the comment if `Run` is supposed to be a blocking call.

> But we could still keep the `BackOff` function as a more graceful best-effort way of throttling the influx of external events when the event queues become big but not too big (some parameters can define what "big" and "too big" is). That way, internal events (e.g. timeouts) would still be processed by the active module and really only the external events (the module inherently knows which events it considers internal and which external) would be stopped, giving the node a chance to also process the "secondary" internal events (results of processing internal events) before the external ones are enabled again. If we do that, I'll reflect that in the comments ofc. Do you see any issues with that approach?

Not sure if we really need this level of control (and complexity), at least until we clearly realize the opposite. Do you have in mind a plausible example of a module that produces both external and internal events?

docs/architecture-decision-records/0003-generalize-modules.md
```diff
@@ -65,23 +68,22 @@ This approach:
         // Note, however, that the implementation may produce output events even without receiving any input.
         // Before Run is called, ActiveModule guarantees not to read or write,
         // respectively, from eventsIn and to eventsOut.
+        //
+        // In general, Run is expected to block until explicitly asked to stop
+	    // (through a module-specific event or by canceling ctx)
```
sergefdrv, line 73:
```suggestion
	// (through a module-specific event or by canceling ctx)
```

docs/architecture-decision-records/0003-generalize-modules.md
```diff
@@ -65,23 +68,22 @@ This approach:
...
-        // This function is necessary for guaranteeing that the internal Node event buffers do not grow indefinitely.
-        // This could potentially happen when external events are injected into the system through active modules
-        // faster than the system can process them.
-        // In such a case, the Node asks all active modules to back off
-        // until enough events in the internal buffers have been processed.
-        BackOff(ctx context.Context)
-        
+        // Note that the node does not guarantee to always read events from eventsOut.
+        // The node might decide at any moment to stop reading from eventsOut for an arbitrary amount of time
+	    // (e.g. if the Node's internal event buffers become full and the Node needs to wait until they free up).
```
sergefdrv, line 79:
```suggestion
	// (e.g. if the Node's internal event buffers become full and the Node needs to wait until they free up).
```

docs/architecture-decision-records/0003-generalize-modules.md
```diff
@@ -65,23 +68,22 @@ This approach:
...
+        //
+        // The implementation of Run is required to always read from eventsIn.
+        // I.e., when an event is written to eventsIn, Run must eventually read it, regardless of the module's state
+        // or any external factors
+        // (e.g., the processing of events must not depend on some other goroutine reading from eventsOut).
```
sergefdrv, line 84: What if the module may require significant time to fully process incoming events (e.g. WAL or application) and other modules keep producing new events for that module?

matejpavlovic: If that module's buffer becomes too big, external events will be put on hold (by stopping to read from active modules' output channels), eventually leading to the other modules also stopping to produce new events (since they always only react to inputs). If modules are properly implemented, this is guaranteed to happen. If they end up "pumping" each other (producing ever-increasing numbers of events in reaction to each other), that's a different issue and definitely a bug in module implementation.

sergefdrv: Ah, that was what you meant by "eventually read".

sergefdrv (approved):

docs/architecture-decision-records/0003-generalize-modules.md
```diff
@@ -0,0 +1,112 @@
...
+        // and return a (potentially empty) list of output events that the application of the input event results in.
+        ApplyEvent(event *eventpb.Event) *events.EventList
+        
+        // Status returns the current state of the module.
+        // Mostly for debugging purposes.
+        Status() (s Status, err error) // The Status type is a placeholder not intended to be defined by this ADR.
+      }
+      ```
+    - `ActiveModule`:
+      Like the `PassiveModule`, the `ActiveModule` consumes input events. However,
```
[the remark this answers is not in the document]
matejpavlovic: Hm my intention / interpretation was the following:

- A `Passive` module does nothing on its own. It only passively sits there, usually does not have any threads, and only *is invoked* by the Node whenever the Node decides. In a way, the `PassiveModule` only contains logic defining how Events should be transformed to other Events, but never "actively" executes it. It is always the Node that executes that logic.

- An `Active` module "actively" pulls and pushes Events from and into the system. It has its own threads that are "actively" running on their own, pull Events from the input queue and push Events into the system. This happens when the module decides, not only when it is asked by the Node. Also when it comes to consuming Events, the `Active` module has a thread that actively consumes them, while the `Passive` module is only waiting for the node to be used for transforming Events.

`Consumer` would be misleading in my view, as it is also associated with "the opposite of producer". And while the module also consumes Events, what distinguishes an `Active` module from a `Passive` one is that in actively produces new Events.

If `PassiveModule` and `ActiveModule` are still misleading, what about `SimpleModule` and `AutonomousModule`? I still have a slight personal preference for `Active`/`Passive` (I was almost proud of myself what nicely fitting names I found for the two kinds of modules 😅 ), but if you prefer `Simple`/`Autonomous`, I'll be happy to switch to that. :) (Or something else, even though `Transformer`/`Consumer` is really misleading to me, so preferably not those)

docs/architecture-decision-records/0003-generalize-modules.md
```diff
@@ -0,0 +1,112 @@
...
+## Considered Options
+
+* **Active and passive modules.**
+  Generalize the concept of modules, such that:
+  - Modules have a unified interface for processing events,
+    where each module internally inspects the incoming events and decides on the appropriate action.
+    The `Protocol` and some other modules already do have such a unified interface - use it for other modules as well.
+  - Instead of dedicated slots for each type of module, use a variable number of generic modules of only 2 types:
+    - `PassiveModule`:
+      A module that simply transforms input events to output events whenever invoked by Mir.
```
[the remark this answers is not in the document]
matejpavlovic: See response to other comment.

PR #89, opened 2022-05-30
Suppress LGTM alerts

sergefdrv:


sergefdrv:
Not sure if this will actually suppress the alerts, so I'll try to merge and see.

PR #98, opened 2022-06-14
Generalize Modules to 2 types: ActiveModule and PassiveModule

matejpavlovic:
This is a very big PR, but each commit should be rather easily digestible.


pkg/iss/iss.go
```diff
@@ -34,6 +35,22 @@ import (
 	t "github.com/filecoin-project/mir/pkg/types"
 )
 
+// ============================================================
+// Global package variables
+// ============================================================
+
+// Names of modules this protocol depends on.
+// The corresponding modules are expected by ISS to be stored under these keys by the Node.
+var (
```
xosmig, line 44: It would be nice for these values to be configurable. Especially `OwnModuleName`, to make it possible to make it possible to run multiple instances of ISS in the same process.

pkg/modules/modules.go
```diff
@@ -10,70 +10,13 @@ SPDX-License-Identifier: Apache-2.0
...
-	App           App              // Implements user application logic. The user is expected to provide this module.
-	WAL           WAL              // Implements a persistent write-ahead log for the case of crashes and restarts.
-	ClientTracker ClientTracker    // Keeps the state related to clients and validates submitted requests.
-	RequestStore  RequestStore     // Provides persistent storage for request data.
-	Protocol      Protocol         // Implements the logic of the distributed protocol.
-	Interceptor   EventInterceptor // Intercepts and logs all internal _Events_ for debugging purposes.
-	Timer         Timer            // Tracks real time (e.g. for timeouts) and injects events accordingly.
+// Module generalizes the ActiveModule and PassiveModule types.
+type Module interface {
+	ImplementsModule()
```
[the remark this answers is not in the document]
sergefdrv: If we do so, the method will be not exported and there will be no way for other packages to directly implement the interface.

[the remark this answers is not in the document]
xosmig: I just checked that it works with the lowercase version. Go uses duck typing and there is no way to prohibit a struct from implementing a method with this name. The visibility only determines who can call the method.

node.go
```diff
@@ -247,161 +280,169 @@ func (n *Node) process(ctx context.Context) error { //nolint:gocyclo
...
+		})
+		selectReactions = append(selectReactions, func(newEventsVal reflect.Value) {
+			newEvents := newEventsVal.Interface().(*events.EventList)
 			if n.debugMode {
-				if n.workChans.debugOut != nil {
-					n.workChans.debugOut <- newEvents
+				if n.debugOut != nil {
+					n.debugOut <- newEvents
 				}
 			} else if err := n.workItems.AddEvents(newEvents); err != nil {
```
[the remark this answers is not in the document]
sergefdrv: Yes, we [already discussed that before](https://github.com/filecoin-project/mir/discussions/51#discussioncomment-2699258).

pkg/modules/modules.go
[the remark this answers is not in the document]
sergefdrv: Otherwise, unexported methods from different packages don't match, see [here](https://blog.chewxy.com/2018/03/18/golang-interfaces/#sealed-interfaces).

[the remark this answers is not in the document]
xosmig: I see. Well, then I want to take my suggestion back: let's keep the uppercase version :)

pkg/iss/iss.go
matejpavlovic: I agree.
Not critical for the moment though. Created an issue for that: #99

matejpavlovic:
I think I addressed all the comments with what I've just pushed. Pls have a look and resolve the conversations if you are happy with the changes. Thanks!

node.go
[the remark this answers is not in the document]
matejpavlovic: Yes, one day we'll need to properly take care of that, but I'd push that for later. Created issue #100 for that.

pkg/modules/eventprocessing.go
```diff
@@ -0,0 +1,138 @@
...
+	for event := iter.Next(); event != nil; event = iter.Next() {
+
+		go func(e *eventpb.Event, j int) {
+
+			// Apply the input event, catching potential panics.
+			res, err := applySafely(e, applyEvent)
+
+			// Wait until the result of processing the previous event has been written to the common channels.
+			if j != 0 {
+				<-done[j-1]
```
xosmig, line 76: Wouldn't it be easier (and, potentially, more efficient) to just have 1 output channel per worker and then read these channels in the right order in the main thread? Then you won't even need the `done` channels.

pkg/modules/eventprocessing.go
```diff
@@ -0,0 +1,138 @@
...
+		done[i] = make(chan struct{})
+	}
+
+	// Start one concurrent worker for each event in the input list.
+	// Note that the processing starts concurrently for all events, and only the writing of the results is synchronized.
+	iter := eventsIn.Iterator()
+	i := 0
+	for event := iter.Next(); event != nil; event = iter.Next() {
+
+		go func(e *eventpb.Event, j int) {
```
xosmig, line 69: I suspect that for optimal performance it may be best to have a bounded number of worker goroutines, but I am not 100% sure and it would probably make the code a bit more complicated.

xosmig:
And what about `Status()`, you decided to keep it out of the `Module` interface?

pkg/clients/signingclienttracker.go
```diff
@@ -115,6 +113,11 @@ func (ct *SigningClientTracker) Status() (s *statuspb.ProtocolStatus, err error)
 	return nil, nil
 }
 
+// The ImplementsModule method only serves the purpose of indicating that this is a Module and must not be called.
+func (ct *SigningClientTracker) ImplementsModule() {
+	panic("ImplementsModule must not be called")
```
xosmig, line 118: The `panic` thing feels a bit excessive to my taste, but if you think that it makes sense to make it explicit in the implementation, then ok.
A single comment in the interface definition could also probably get the job done :)

pkg/modules/eventprocessing.go
matejpavlovic: Yes, I was thinking about that too, but for now I think what we have is good enough. From experience, the performance difference was marginal / not observable in an analogous case in a previous project. If this becomes an issue, we can easily limit the number of threads later.

matejpavlovic:
> And what about `Status()`? Have you decided to keep it out of the `Module` interface?

Oh yes, I forgot about that. Fixed now.

pkg/modules/eventprocessing.go
```diff
@@ -0,0 +1,122 @@
...
+	errors := make([]chan error, eventsIn.Len())
+	for i := 0; i < eventsIn.Len(); i++ {
+		results[i] = make(chan *events.EventList)
+		errors[i] = make(chan error)
+	}
+
+	// Start one concurrent worker for each event in the input list.
+	// Note that the processing starts concurrently for all events, and only the writing of the results is synchronized.
+	iter := eventsIn.Iterator()
+	i := 0
```
xosmig, line 58: I think this variable is no longer used.

pkg/clients/signingclienttracker.go
matejpavlovic: Yes I was hesitating too, but in the end I kept it for completeness sake. The comment in the definition is now there too.

pkg/modules/eventprocessing.go
matejpavlovic: The `i` variable? It is, as each goroutine needs to know the index of its output channels and the `Iterator` implementation does not provide that information.

xosmig:
LGTM

pkg/modules/modules.go
```diff
@@ -10,12 +10,20 @@ SPDX-License-Identifier: Apache-2.0
...
 // Module generalizes the ActiveModule and PassiveModule types.
 type Module interface {
+
+	// ImplementsModule only serves the purpose of indicating that this is a Module and must not be called.
 	ImplementsModule()
+
+	// Status returns the current state of the module.
+	// Mostly for debugging purposes.
+	// TODO: This functionality is not yet implemented and all the Status() implementations are stubs. Fix that.
+	Status() (s *statuspb.ProtocolStatus, err error)
```
sergefdrv, line 26: This method is never used nor implemented in the current code. Maybe just remove it for now?

sergefdrv (commented):
That is a massive PR! Good job :+1:

pkg/simplewal/simplewal.go
```diff
@@ -1,9 +1,3 @@
-/*
-Copyright IBM Corp. All Rights Reserved.
-
-SPDX-License-Identifier: Apache-2.0
-*/
-
```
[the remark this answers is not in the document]
matejpavlovic: I guess it is. I had a tendency to leave it in files that did not fundamentally change since my time at IBM, but now basically all the files contain a mix of IBM and non-IBM code. Will remove.

pkg/iss/iss.go
```diff
@@ -34,6 +35,16 @@ import (
...
+	OwnModuleName t.ModuleID = "iss"
```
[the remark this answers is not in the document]
matejpavlovic: No this variable (and the others added in future commits) don't need to be exported.
Will fix.

node.go
```diff
@@ -422,33 +422,69 @@ func (n *Node) startModules(ctx context.Context, wg *sync.WaitGroup) {
...
+		case newEvents, ok := <-eventsIn:
+			if ok {
+				select {
+				case n.workChans.workItemInput <- newEvents:
+				case <-ctx.Done():
+					return
+				case <-n.workErrNotifier.ExitC():
+					return
+				}
+			}
```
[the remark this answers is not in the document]
matejpavlovic: Good catch! Will fix in place.

pkg/deploytest/deployment.go
```diff
@@ -252,7 +252,23 @@ func (d *Deployment) Run(ctx context.Context) (finalStatuses []NodeStatus, heapO
...
+	case <-func() chan struct{} {
+		ch := make(chan struct{})
+		go func() {
+			nodeWg.Wait()
+			close(ch)
+		}()
+		return ch
+	}():
+
+	}
```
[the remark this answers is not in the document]
matejpavlovic: Yes this is more readable. Will fix.

pkg/modules/modules.go
matejpavlovic: In fact I've been hesitating to do that for some time already and kept it around because it has been part of the old design. But there is no rational reason to keep it at this point, we'll re-introduce it when we are ready to implement it. Will remove.

pkg/clients/signingclienttracker.go
matejpavlovic: I'd like to keep the comments. As a reader of that code I'm sure I'd appreciate them.
The panics probably are a bit over the top indeed, I'll remove them.

pkg/deploytest/fakeapp.go
```diff
@@ -9,13 +9,18 @@ package deploytest
...
+	"github.com/filecoin-project/mir/pkg/pb/statuspb"
+	t "github.com/filecoin-project/mir/pkg/types"
 
 	"github.com/filecoin-project/mir/pkg/modules"
 	"github.com/filecoin-project/mir/pkg/pb/requestpb"
 )
 
 // FakeApp represents a dummy stub application used for testing only.
 type FakeApp struct {
+	modules.PassiveModule
```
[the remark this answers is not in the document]
matejpavlovic: Actually, your comment comes from inspecting an early commit. It has already been addressed in a [later commit](https://github.com/filecoin-project/mir/pull/98/commits/3918c9042fb5b79d0c64f4f34f546ecfd63e3db7)

protos/eventpb/eventpb.proto
```diff
@@ -59,6 +59,8 @@ message Event {
   // In this case, the WAL append event would be this event
   // and the next field would contain the message sending event.
   repeated Event next = 100;
+
+  string dest_module = 200;
```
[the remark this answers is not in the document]
sergefdrv: I think it could be set for all events by the Node itself. This way all events would have both the recipient and the sender. Whether to use this information not would be up to the event consumer. But it's just a suggestion to unify code, it's up to you.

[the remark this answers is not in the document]
xosmig: This would (slightly) complicate the interface of the `DslModule` because all handlers would have to accept the source as a separate parameter.

A bit of context: `DslModule` is a construction that I want to propose to add, which will allow writing sequential passive modules in a more declarative pseudocode-like way. I will make a PR with it shortly after this PR is merged.

pkg/deploytest/fakeapp.go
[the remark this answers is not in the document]
sergefdrv: Oh, I just reviewed the PR commit-by-commit...

pkg/simplewal/simplewal.go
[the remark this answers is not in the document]
matejpavlovic: Yes it is.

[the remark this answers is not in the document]
matejpavlovic: Actually, just realized that I got it all wrong, thinking you were suggesting to remove it, while you were suggesting to keep it. I must be color blind... 🙄 (I don't remember removing it and it must have happened by accident, probably when copying code from another version of the file. That's why I assumed you suggesting removal.)
Let's keep it then, to be on the safe side.

matejpavlovic:
Ok I think I addressed all the stuff from above that required changes, plus I fixed the problem with the debugger not working.
Please let me know if I forgot anything, otherwise pls resolve the conversations so I can merge the PR. Thanks both for the improvements!

pkg/clients/signingclienttracker.go
```diff
@@ -1,29 +1,26 @@
-/*
-Copyright IBM Corp. All Rights Reserved.
-
-SPDX-License-Identifier: Apache-2.0
-*/
-
```
sergefdrv, line 6: Keep it?

pkg/crypto/api.go
```diff
@@ -1,20 +1,15 @@
-/*
-Copyright IBM Corp. All Rights Reserved.
-
-SPDX-License-Identifier: Apache-2.0
-*/
-
```
sergefdrv, line 6: Keep it?

pkg/iss/iss.go
```diff
@@ -29,11 +30,26 @@ import (
...
+// The corresponding modules are expected by ISS to be stored under these keys by the Node.
+var (
+	ownModuleName    t.ModuleID = "iss"
+	netModuleName    t.ModuleID = "net"
+	appModuleName    t.ModuleID = "app"
+	walModuleName    t.ModuleID = "wal"
+	hasherModuleName t.ModuleID = "hasher"
+	cryptoModuleName t.ModuleID = "crypto"
+	timerModuleName  t.ModuleID = "timer"
+)
```
sergefdrv, line 51: Those string literals still appear here and there. Maybe though make these constants exported and use throughout the code base consistently?

pkg/iss/iss.go
```diff
@@ -29,11 +30,26 @@ import (
...
 )
 
+// ============================================================
+// Global package variables
+// ============================================================
+
+// Names of modules this protocol depends on.
+// The corresponding modules are expected by ISS to be stored under these keys by the Node.
+var (
+	ownModuleName    t.ModuleID = "iss"
```
sergefdrv, line 44:
```suggestion
	ISSModuleName    t.ModuleID = "iss"
```

pkg/simplewal/simplewal.go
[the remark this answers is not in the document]
sergefdrv: Yes, I think it is better to keep the original copyright statement.

pkg/clients/signingclienttracker.go
matejpavlovic: The content of this file changed almost completely since IBM, I think here we can remove it without problems.

pkg/crypto/api.go
matejpavlovic: This file is also very recent, the content already created outside of IBM, so it can stay removed.

pkg/iss/iss.go
matejpavlovic: The literals are mostly used in modules that are going to be phased out for ISS, it's not worth the effort at this point. When the module composition becomes stable, we can consolidate much easier.

xosmig: I think we should avoid global constants and allow, for example, different modules to use different hash implementations. For now, these names are hard-coded for ISS, but in the future, I hope they will come from a config.

matejpavlovic: Yes, I already created an issue (#99) for that yesterday.

pkg/clients/signingclienttracker.go
sergefdrv: https://github.com/filecoin-project/mir/compare/10658fbe24ca2bb33ac3151daf7d6f610f4a16e2...matejpavlovic:general-modules#diff-c9a08836d7ab42d845d17eddc4a99521169a6c945cb85923a8f774c1e739c85d

pkg/crypto/api.go
matejpavlovic: In fact, I've slimmed it down even more (first I had it on a different branch, but in fact it also falls into "module generalization", as it generalizes the interface of what used to be the crypto module). Pushing the new version here.

sergefdrv: [This file is almost the same](https://github.com/filecoin-project/mir/compare/10658fbe24ca2bb33ac3151daf7d6f610f4a16e2...matejpavlovic:general-modules#diff-3452a5463a9ec45fcb2cd44aaee2013b82bdba6c182cd2d1da6c159da4bb5085).

pkg/clients/signingclienttracker.go
matejpavlovic: Hm I thought I wrote it much more recently. Fair enough, let's keep the notice in this one.

pkg/crypto/api.go
matejpavlovic: Wow ok, I really thought I wrote it after leaving IBM. Let's keep it then.

xosmig: Btw, apparently, formally, you should indicate that the files have been changed.

> You must cause any modified files to carry prominent notices stating that You changed the files.

pkg/iss/iss.go
matejpavlovic: But, unless I overlooked something (possible), the literals don't appear *within* the iss package. The other occurrences fall in the scope of my [previous comment](https://github.com/filecoin-project/mir/pull/98#discussion_r898951057):
> The literals are mostly used in modules that are going to be phased out for ISS, it's not worth the effort at this point. When the module composition becomes stable, we can consolidate much easier.

PR #102, opened 2022-06-17
[WIP] DSL interface for high-level modules

xosmig:
The implementation itself (file `pkg/modules/dsl/dslmodule.go`) is mostly ready albeit not yet tested, but the work on the motivating example (file `pkg/cb/cbmodule.go`) is still in progress.

I suggest first reading the example (file `pkg/cb/cbmodule.go`) to understand the main idea, then the core implementation (file `pkg/modules/dsl/dslmodule.go`), and only then the boilerplate (files `pkg/modules/dsl/events.go`, `pkg/cb/cbdsl/dslmodule.go`, `pkg/cb/cbdsl/events.go`, and the rest).


sergefdrv:
@xosmig This is an interesting approach to using Mir framework. Apart from the usual event-based notation, I would also like to explore something like the control-oriented notation mentioned in [this](https://youtu.be/jP1Mfh6Qkzw?t=486) lecture, but I'm not sure if that is easy to do in Mir architecture.

xosmig:
@sergefdrv if I understand you correctly, this is similar to [coroutines](https://en.wikipedia.org/wiki/Coroutine) being executed on 1 thread (i.e., without any parallelism or preemption). For pseudocode, I also prefer this notation over event-based and often use it. Using it in an actual implementation may be a bit trickier because the control flow is less explicit. 

Note that this is rather an extension than a replacement to the event-based notation.

Here is a slightly bigger example of pseudocode using this notation than the one in the video (from [this paper](https://arxiv.org/pdf/2005.13499.pdf)):

image

Note that it uses both "wait for" and "upon" constructions. If converted to actual code, it would probably need a few more "upon" handlers to manage the replies for lines 179 and 185.

It does seem to be doable in mir: we will need to run a separate goroutine each time we execute a handler. Then the main goroutine would act as a coordinator to make sure that at most 1 goroutine at a time is actually being executed. When the handler reaches a waiting statement, it tells the main goroutine the event type it is waiting for and blocks until it gets the event back from the main goroutine. The main goroutine, in turn, blocks until the handler tells it that it either has reached a waiting command or has finished.

The downside of this implementation is that it incurs the overhead of creating a goroutine for each event handler even if you don't use the "wait for" statement. Moreover, in the current design of dsl modules, each event may have multiple handlers (hence, there would be multiple goroutines each time an event arrives).

Here are some other ways to implement it I could think of. None of them seem to be practical:

  1. Use 1 goroutine. Manually save the goroutine context (the state of the CPU registers and so on) on a waiting statement and then manually restore it. However, I'm not sure if go runtime exports an API that would allow us to do it (I couldn't find anything similar in the [go runtime package](https://pkg.go.dev/runtime)). If there was a coroutine library for go, we could probably use it, but I couldn't find any (maybe it's a sign that it is in fact impossible in go).

  2. Construct (or extract) the [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree) of the (pseudo)code in runtime and then interpret them in runtime. This would probably be very slow. Potential ways to get the AST: 
    i. Use the [parser](https://pkg.go.dev/go/parser) package.
    ii. Replace all language operations by custom functions: instead of writing `x = y`, you would write `ast.Assign(x, y)`, which would create a node in the AST for this assignment.  
    iii. (not applicable to go) In a language like C++, we could create fake objects, which would override operations like `=` to create AST nodes instead of actually doing the assignment.

  3. Finally, there is always an option to use `go generate` to generate event-based code from the control-oriented (pseudo)code.

xosmig:
Perhaps, the overhead could be lowered. Instead of creating a goroutine for each event handler, the main goroutine can create 1 worker goroutine and pass event handlers to it one by one. After passing a handler to the worker, the main goroutine waits until the worker tells it that it either has finished processing the handler or that it blocked on a "wait for" statement. In the latter case, the main goroutine creates a new worker and continues processing events using it.

In this case, goroutines will be created only when it is actually necessary.

xosmig:
To optimize even further, the main goroutine can just pass the whole list of events to the worker. The worker executes them until it encounters a "wait for" statement, in which case it passes the tail of the list to the main goroutine and blocks.

matejpavlovic:
Both ideas (DSL moudle and coroutines) are a very interesting further development of Mir that I'm all for pursuing.
AFAIU, the DSL module is a really nice add-on that seems basically implemented at this point and can directly be very useful for protocol implementation. The coroutine extension seems also very useful, but seems to involve much more implementation overhead if we wanted to implement it now.

Thus, I'd suggest moving the coroutine discussion to a separate issue and continuing to discuss there, so the ideas don't get lost. At this point, I think we should focus on the most immediate goal of implementing the basic availability layer (basic Narwhal / multisig certificate collector protoco) so we can reach the goal for our milestone that is coming closer and closer. It seems to me that the DSL module will probably accelerate this, as it seems almost ready. Then we can pick up the coroutines again and see whether to also implement them.

This is definitely not meant to kill the discussion on the coroutine-based programming model. On the contrary, I'm encouraging that, also in form of such spontaneous brainstorms. Just want to be on the same page about the immediate goals in terms of writing code :)

sergefdrv:
> similar to coroutines being executed on 1 thread (i.e., without any parallelism or preemption)

Not necessarily, I think there are many way to achieve this. I was actually considering something like communicating sequential processes. One could also do waiting with traditional mutexes and conditional variables, or Go's select statements, or STM, or whatever...

> It does seem to be doable in mir

I wouldn't try sticking to Mir architecture when exploring possible ways to implement the alternative approach, but I don't have any specific idea at this point. Clearly, this is out of the scope for now due to immediate higher-priority tasks.

> coroutine-based programming model

I'd still like to put focus more generally: on control-oriented notation.

pkg/cb/cbdsl/events.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/filecoin-project/mir/pkg/modules/dsl"
+	"github.com/filecoin-project/mir/pkg/pb/cbpb"
+	"github.com/filecoin-project/mir/pkg/pb/eventpb"
+	"github.com/filecoin-project/mir/pkg/pb/messagepb"
+	t "github.com/filecoin-project/mir/pkg/types"
+)
+
+// Contains event wrappers specific for this module
+
+func SignRequest(m *cbModuleImpl, destModule t.ModuleID, data [][]byte) {
```
matejpavlovic, line 13: This might be a matter of personal taste, but it could be more intuitive to have this function as well as the other ones in this file be methods of m.

matejpavlovic (commented):
Looks nice, just some minor comments. Didn't look through the context store, as I assume it to be the same as in #103

pkg/modules/dsl/dslmodule.go
```diff
@@ -0,0 +1,130 @@
...
+// dslModuleImpl allows creating passive modules in a very natural declarative way.
+type dslModuleImpl struct {
+	eventHandlers     map[reflect.Type][]func(ev *eventpb.Event) error
+	conditionHandlers []func() error
+	outputEvents      *events.EventList
+}
+
+type Handle *dslModuleImpl
+
+type DslModule interface {
```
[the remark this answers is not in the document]
xosmig: Yes, it does, in order to make it possible to extend it (see example in `pkg/cb/cbdsl/dslmodule.go`). If DslModule was a struct, then functions that accept `*DslModule` wouldn't accept `*cbdsl.cbModuleImpl` (go doesn't do implicit cast to an embedded type).

pkg/cb/cbdsl/events.go
xosmig: The examples aren't ideal because they operate on `cbdsl.cbModuleImpl` and can be methods of this type. However, if there is a package that defines a new event type (assuming one day we will make it possible to define event types outside the framework itself), then the `UponEvent(m, ...)`  and `Event(m, ...)` should be defined for general `DslModule` and in that package. This would not be possible for methods.

xosmig: Aesthetically, I would also prefer using methods :)

xosmig:
> Didn't look through the context store, as I assume it to be the same as in https://github.com/filecoin-project/mir/pull/103

Oh, I think I committed it here by accident. Thank you for noticing.

xosmig:
@sergefdrv sorry, apparently, I didn't quite get what you mean by "control-oriented notation". The example in the video looks similar to the model I mentioned, but, apparently, it's a coincidence.

xosmig:
> At this point, I think we should focus on the most immediate goal of implementing the basic availability layer

Yes, sure :)
I didn't really consider implementing the "wait for" statement in this PR or in the near future. It looks more like a nice-to-have but an unnecessary feature.

As for control-oriented notation, it seems to be a big separate direction.

xosmig:
Two major changes to the DSL module design:

1. Introduced a standardized way for context management for asynchronous operations like signature verification and hash computation.
    * **pros:** writing code that uses such asynchronous functions now require much less boilerplate.
    * **cons:** none? :)
    * I also had a couple of alternative designs for managing the context. I chose this one as it was the simplest.

2. Replaced `UponEvent[EvTp, Ev](m Module, func(ev *Ev) error)` by `RegisterEventHandler[EvTp](m, func(ev *EvTp) error)`.
    * **a bit of context:** EvTp  is the generated "wrapper" type (e.g., `eventpb.Event_Request`) and Ev  is the actual event type, (e.g., `eventpb.Request`). `EvTp` is structurally identical to `struct { ev *Ev }` (unless `Ev` is a primitive type, in which case there is no pointer).
    * **pros:** due to some quirks of the code generated by the protobuf compiler, `UponEvent` had to verify that `EvTp` and `Ev` are compatible in runtime (at the moment of handler registration). Moreover, the object of type `Ev` was extracted from the object of type `EvTp` in a rather hacky way (see [this file](https://github.com/xosmig/mir/blob/5ae3238898c40fb01d219e4c6168182f53980531/pkg/util/protoutil/protoutil.go)).
    * **cons:** `RegisterEventHandler` is much less user-friendly and goes against the "declarative" nature of dsl modules.
    * **motivation:** in the initial design, `UponEvent` was supposed to be actually called in the protocol implementation. However, the design has evolved, and now the protocol implementation is not really supposed to use this function directly and is rather supposed to use the wrappers like the ones in `pkg/dsl/events.go`. For the wrappers (which are separated from the protocol logic), it is ok to use a slightly less user-friendly and non-declarative function for the sake of type safety.

xosmig:
I think the design is more or less stable at this point. I've started adding some unit tests.
@matejpavlovic @sergefdrv it would be nice if you could do a high-level review before I invest too much time in polishing the current code.

pkg/bcb/protobuf.go
```diff
@@ -0,0 +1,51 @@
...
+			},
+		},
+	})
+}
+
+func FinalMessage(moduleID t.ModuleID, data []byte, signers []t.NodeID, signatures [][]byte) *messagepb.Message {
+	var signersPb []string
+	for _, signer := range signers {
+		signersPb = append(signersPb, signer.Pb())
+	}
```
matejpavlovic, line 40: You can use `t.NodeIDSlicePb(signers)` instead of this.

pkg/dsl/dslmodule.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+type dslModuleImpl struct {
+	moduleID          t.ModuleID
+	eventHandlers     map[reflect.Type][]func(ev *eventpb.Event) error
+	conditionHandlers []func() error
+	outputEvents      *events.EventList
+	// contextStore is used to store and recover context on asynchronous operations such as signature verification.
+	contextStore cs.ContextStore[any]
+	// eventCleanupContextIDs is used to dispose the
```
matejpavlovic, line 21: dispose the...?

pkg/dsl/dslmodule.go
```diff
@@ -0,0 +1,153 @@
...
+// Module allows creating passive modules in a very natural declarative way.
+type Module interface {
+	modules.PassiveModule
+
+	// GetDslHandle returns an object that
+	GetDslHandle() Handle
+
+	// GetModuleID returns the identifier of the module.
+	// TODO: consider moving this method to modules.Module.
+	GetModuleID() t.ModuleID
```
matejpavlovic, line 40: Just a cosmetic suggestion: remove the `Get` (more idiomatic Go and nicer to read)

pkg/dsl/dslmodule.go
```diff
@@ -0,0 +1,153 @@
...
+	m.GetDslHandle().impl.eventHandlers[evTpPtrType] = append(m.GetDslHandle().impl.eventHandlers[evTpPtrType],
+		func(event *eventpb.Event) error {
+			evTpPtr := ((any)(event.Type)).(*EvTp)
+			return handler(evTpPtr)
+		})
+}
+
+// UponCondition registers a special type of handler that will be invoked each time after processing a batch of events.
+// The handler is assumed to represent a conditional action: it is supposed to check some predicate on the state
+// and perform actions if the predicate evaluates is satisfied.
```
matejpavlovic, line 77: Typo "evaluates is satisfied"

pkg/dsl/dslmodule.go
```diff
@@ -0,0 +1,153 @@
...
+	impl.conditionHandlers = append(impl.conditionHandlers, handler)
+}
+
+// StoreContext stores the given data and returns an automatically deterministically generated unique id.
+// The data can be later recovered or disposed of using this id.
+func (h Handle) StoreContext(context any) ContextID {
+	return h.impl.contextStore.Store(context)
+}
+
+// CleanupContext schedules a disposal of context with the given id after the current batch of events is processed.
```
matejpavlovic, line 89: What is the reason (I'm sure there is one) for not just disposing of the context right away?

pkg/dsl/dslmodule.go
```diff
@@ -0,0 +1,153 @@
...
+func (h Handle) CleanupContext(id ContextID) {
+	h.impl.eventCleanupContextIDs[id] = struct{}{}
+}
+
+// RecoverAndRetainContext recovers the context with the given id and retains it in the internal context store so that
+// it can be recovered again later. Only use this function when expect to receive multiple events with the same context.
```
matejpavlovic, line 95: when expecting

matejpavlovic (approved):
Very nice, good job!

pkg/dsl/dslmodule.go
xosmig: One of the central design choices is that each event can be processed by multiple handlers.
So, you cannot dispose of the context after it is processed by just one handler. You could do it after the whole event is processed, but I decided that the semantics would be clearer if we do it after the whole batch is processed.

matejpavlovic: Makes sense.

pkg/dsl/dslmodule.go
xosmig: Fixed.

pkg/dsl/dslmodule.go
xosmig: Fixed.

pkg/bcb/protobuf.go
xosmig: Nice! Fixed.

pkg/dsl/dslmodule.go
xosmig: Oh, sorry. Fixed.

pkg/dsl/dslmodule.go
xosmig: I added on that.

xosmig:
@matejpavlovic thanks for your review! Could you please also take a look at [this commit](https://github.com/filecoin-project/mir/pull/102/commits/bbac2f1bdab7fa80f25127b5b16c3c57b5e63379)? Do I interpret the `NodeSigsVerified` message correctly?

The rest of the commits after your review are tests and documentation.

pkg/dsl/events.go
```diff
@@ -138,26 +146,30 @@ func UponNodeSigsVerified[C any](
...
-			nodeIds = append(nodeIds, t.NodeID(id))
+		nodeIds := make([]t.NodeID, len(ev.NodeIds))
+		for i := range ev.NodeIds {
+			nodeIds[i] = t.NodeID(ev.NodeIds[i])
 		}
 
-		var errs []error
-		for _, err := range ev.Errors {
-			errs = append(errs, errors.New(err))
+		errs := make([]error, len(ev.Valid))
```
matejpavlovic, line 149: Is this copying intended for better isolation?
Otherwise, the `ev.Errors` slice could be used directly.
If you do want a copy and errors are actually values and not pointers (not 100%), you can also simply use `copy` to create a new error slice instead of looping through the errors.

pkg/dsl/events.go
```diff
@@ -138,26 +146,30 @@ func UponNodeSigsVerified[C any](
 			return nil
 		}
 
-		var nodeIds []t.NodeID
-		for _, id := range ev.NodeIds {
-			nodeIds = append(nodeIds, t.NodeID(id))
+		nodeIds := make([]t.NodeID, len(ev.NodeIds))
```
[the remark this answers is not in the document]
xosmig: Already fixed in a [later commit](https://github.com/filecoin-project/mir/pull/102/commits/de28cda034f354c06fb1ab43cf441b3cae88e3eb).

pkg/dsl/events.go
xosmig: It's for the sake of type conversion. `ev.Errors` is a slice of strings and I need a slice of errors (with nil indicating that no error). Note that the handler doesn't accept the `ev.Valid` slice. In some sense, the information from `ev.Valid []bool` and `ev.Errors []string` is "merged" into `errs []error`, to make it more idiomatic go.

matejpavlovic: Right, makes sense. In fact, I've been hesitating removing the slice of `bool`s anyway. As you probably noticed, the `allOk` field is also just sugar. But that seems to be rather convenient, so I'm keeping it there too.

PR #105, opened 2022-06-20
Simplifications

matejpavlovic:
This PR simplifies both Mir and its ISS implementation by
- generalizing the node-level event injection interface
- removing all the availability-related logic from ISS
- getting rid of client request authentication and related code.


matejpavlovic:
This PR is mostly about deleting legacy code and adapting the rest accordingly.
The interface does change, however (to a simpler and more general one) for the `Node` and `Crypto` - this concerns @dnkolegov in particular at this moment.

dnkolegov (approved):

PR #109, opened 2022-06-22
Enhance slow-propose test configuration

sergefdrv:


pkg/deploytest/deployment.go
```diff
@@ -128,13 +128,19 @@ func NewDeployment(testConfig *TestConfig) (*Deployment, error) {
...
-		if i == 0 {
-			issConfig = testConfig.FirstReplicaISSConfig
+		// ISS configuration
+		issConfig := iss.DefaultConfig(membership)
+		if testConfig.SlowProposeReplicas[i] {
+			// Increase MaxProposeDelay such that it is likely to trigger view change by segment timeout
+			issConfig.MaxProposeDelay = (issConfig.PBFTViewChangeSegmentTimeout + issConfig.MaxProposeDelay) / time.Duration(issConfig.SegmentLength)
+
+			// Make sure it also has change to trigger view change by batch timeout
+			if issConfig.MaxProposeDelay < issConfig.PBFTViewChangeBatchTimeout {
```
matejpavlovic, line 138: In fact, doesn't this obviate the previous lines? At the end of the day, it is always the batch timeout that is triggered, making the segment timeout irrelevant. It looks like simply always setting `MaxProposeDelay` to the batch timeout has the same effect in practice.
One could argue that if the `MaxProposeDelay` is *exactly* the same as the batch timeout, the view change might or might not be triggered, but in practice, it almost always does, making the segment timeout extremely unlikely, especially with segments that are not trivially short.
Bottom line: what about just
```go
if testConfig.SlowProposeReplicas[i] {
    issConfig.MaxProposeDelay = issConfig.PBFTViewChangeBatchTimeout
}
```

pkg/deploytest/deployment.go
```diff
@@ -128,13 +128,19 @@ func NewDeployment(testConfig *TestConfig) (*Deployment, error) {
...
-		var issConfig *iss.Config
-		if i == 0 {
-			issConfig = testConfig.FirstReplicaISSConfig
+		// ISS configuration
+		issConfig := iss.DefaultConfig(membership)
+		if testConfig.SlowProposeReplicas[i] {
+			// Increase MaxProposeDelay such that it is likely to trigger view change by segment timeout
+			issConfig.MaxProposeDelay = (issConfig.PBFTViewChangeSegmentTimeout + issConfig.MaxProposeDelay) / time.Duration(issConfig.SegmentLength)
+
+			// Make sure it also has change to trigger view change by batch timeout
```
matejpavlovic, line 137: Typo

pkg/deploytest/deployment.go
sergefdrv: The logic was to make it trigger the segment timeout, but ensure that batch timeout can also be triggered, without making any assumption about the default timeout values.

matejpavlovic: But the way it is now, it almost never triggers the segment timeout in practice, only the batch timeout.

sergefdrv: It happens that the following holds in the current default configuration: 
```
PBFTViewChangeSegmentTimeout < SegmentLength * PBFTViewChangeBatchTimeout
```
but I didn't want to assume this.

sergefdrv: > One could argue that if the `MaxProposeDelay` is exactly the same as the batch timeout, the view change might or might not be triggered, but in practice, it almost always does, making the segment timeout extremely unlikely, especially with segments that are not trivially short.

This may not be necessarily true when we switch into using simulated network and local clocks.

matejpavlovic: If that didn't hold, the segment timeout would become almost irrelevant. In such a case, if a node didn't commit everything by the segment timeout, it inevitably must have observed at least one batch timeout and already entered view change (modulo, perhaps, some corner cases around segment-level checkpoints).

sergefdrv: > if a node didn't commit everything by the segment timeout, it inevitably must have observed at least one batch timeout

Oh, that observation actually makes a good case for setting `MaxProposeDelay = PBFTViewChangeBatchTimeout` :+1:

sergefdrv:
Summary of [changes](https://github.com/filecoin-project/mir/compare/5921ee6d13b2e111261e43eef492c545e13f5de7..9642503b429db150505db6cc536e98761a390156): simplified slow config to setting `MaxProposeDelay = PBFTViewChangeBatchTimeout`

matejpavlovic (approved):
👍

PR #111, opened 2022-06-22
Replace Ginkgo and Gomega with Testify

sergefdrv:
Some of the reasons to prefer native Go testing with Testify:
 - Ginkgo doesn't seem very convenient to use, e.g. running an
   arbitrary subset of tests
 - Ginkgo doesn't seem to have native integration with Go's benchmark
   or fuzz testing facilities


matejpavlovic (approved):
👍

PR #113, opened 2022-06-23
PBFT instance-level checkpoints

matejpavlovic:
If correct nodes stop participating in the SB instance protocol immediately after having delivered all batches in a segment,
a minority of nodes might get stuck in higher views if they did not deliver all batches yet,
never finding enough support for finishing a view change.
The high-level checkpoints (encompassing all segments) are not enough to resolve this problem,
because multiple segments can block each other, each with its own majority of nodes that finish it,
but with too few nodes that finished all segments.
If, for example, each segment only has one node that fails to complete it, no high-level checkpoint can be constructed,
as too few nodes will have delivered everything from all the segments.
Thus, local instance-level checkpoints are required, so nodes can catch up on each segment separately.


cmd/mircat/debug.go
```diff
@@ -19,6 +16,8 @@ import (
 	"github.com/filecoin-project/mir/pkg/pb/recordingpb"
 	t "github.com/filecoin-project/mir/pkg/types"
 	"google.golang.org/protobuf/encoding/protojson"
+	"io"
+	"os"
```
sergefdrv, line 20: Why moving these imports here?

pkg/iss/pbftsegmentchkp.go
```diff
@@ -0,0 +1,302 @@
...
+		return
+	}
+
+	// Store Done message
+	chkp.doneMessages[nodeID] = doneMsg
+	strKey := aggregateToString(doneMsg.Digests)
+	chkp.doneMsgIndex[strKey] = append(chkp.doneMsgIndex[strKey], nodeID)
+
+	// If a quorum of nodes has sent a Done message
+	if nodeIDs := chkp.consistent(len(segment.Membership)); nodeIDs != nil {
```
sergefdrv, line 97: Why do we go over the whole `chkp.doneMsgIndex` inside `chkp.consistent` if we know that only `chkp.doneMsgIndex[strKey]` has changed?

pkg/iss/pbftsegmentchkp.go
```diff
@@ -0,0 +1,302 @@
...
+	if pbft.allCommitted() {
+		eventsOut.PushBackList(pbft.sendDoneMessages())
+	}
+
+	// Deliver batch.
+	eventsOut.PushBack(pbft.eventService.SBEvent(SBDeliverEvent(
+		sn,
+		slot.Preprepare.Batch,
+		slot.Preprepare.Aborted,
+	)))
```
sergefdrv, line 299: Do we want to emit `SBDeliver` event even if the digest does not match?

pkg/iss/pbftsegmentchkp.go
matejpavlovic: Good point :-) Fixed.

pkg/iss/pbftsegmentchkp.go
matejpavlovic: Very good catch! Fiexed.

pkg/iss/iss.go
```diff
@@ -776,7 +761,8 @@ func (iss *ISS) initEpoch(newEpoch t.EpochNr) {
 
 	// Set the new epoch number and re-initialize list of orderers.
 	epoch := &epochInfo{
-		Nr: newEpoch,
+		Nr:         newEpoch,
+		Membership: iss.config.Membership, // TODO: Make a proper copy once reconfiguration is supported.
```
sergefdrv, line 765: > TODO: Make a proper copy once reconfiguration is supported

Do you plan to recycle the same ISS module instance for multiple configurations?

cmd/mircat/debug.go
sergefdrv: Hmm, my editor automatically separates built-in imports imports from normal ones since `gofmt -s` doesn't like the other way.

pkg/iss/pbftslot.go
```diff
@@ -93,14 +93,23 @@ func (slot *pbftSlot) advanceState(pbft *pbftInstance, sn t.SeqNr) *events.Event
...
-			eventsOut.PushBack(pbft.eventService.TimerDelay(
-				t.TimeDuration(pbft.config.ViewChangeBatchTimeout),
-				pbft.eventService.SBEvent(PbftViewChangeBatchTimeout(pbft.view, pbft.numCommitted(pbft.view))),
-			))
+		// It will be ignored if any of those values change by the time the timer fires
+		// or if a quorum of nodes confirms having committed all batches.
+		eventsOut.PushBack(pbft.eventService.TimerDelay(
+			t.TimeDuration(pbft.config.ViewChangeBatchTimeout),
+			pbft.eventService.SBEvent(PbftViewChangeBatchTimeout(pbft.view, pbft.numCommitted(pbft.view))),
+		))
```
sergefdrv, line 106: If we have now the end-of-segment checkpoints then why do we start the batch timeout even though all the slots are committed?

pkg/iss/iss.go
matejpavlovic: Potentially, haven't decided yet. I think both options are viable and I'll see what is better only when actually writing the code implementing it. Since ISS itself is modular and basically only creates and destroys instances of sub-protocols, it might make sense to keep the same ISS module, especially since it keeps around data structures that are membership-independent (e.g. log of committed batches, request buckets). If it turns out to be simpler to completely delete the ISS module and create a new one, I'll do that.

For the time being, I'd keep this comment until the decision is made.

pkg/iss/pbftslot.go
matejpavlovic: Because all the slots might be committed only locally. We need the timeout for the situation where only a few nodes committed all slots locally, but not enough to have a stable segment-level checkpoint. In that case, even those that committed all the slots are necessary for starting a view change protocol that the other correct nodes need. Only when the segment-level checkpoint is stable can we start ignoring timers.

That said, technically, it would not be strictly necessary, since there is still the end-of-segment timeout, but continuing with the batch timer seems rather natural to me.

sergefdrv: How about the following?

```go
if !pbft.allCommitted() || !pbft.segmentCheckpoint.Stable(len(pbft.segment.Membership)) {
	eventsOut.PushBack(pbft.eventService.TimerDelay(
		t.TimeDuration(pbft.config.ViewChangeBatchTimeout),
		pbft.eventService.SBEvent(PbftViewChangeBatchTimeout(pbft.view, pbft.numCommitted(pbft.view))),
	))
}
```

matejpavlovic: Technically that check is redundant if one looks purely at correctness, since it still must be repeated when the timer fires. Checking here is an optimization to potentially save ourselves a timer event that we already know will be ignored. But it probably also improves the understandability of the code, showing explicitly that we don't need the timeout if the checkpoint is stable.

Bottom line: I added the check :-)
(I also updated the semantics of "stability" of a segment-level checkpoint to be consistent with iss-level checkpoints and require not only having received a quorum of Done messages, but also having committed all slots. That's why I'm now only checking for checkpoint stability.)

pkg/iss/config.go
```diff
@@ -88,6 +88,13 @@ type Config struct {
 	// Must not be negative.
 	MsgBufCapacity int
 
+	// Number of most recent epochs that are older than the latest stable checkpoint.
+	RetainedEpochs int
```
sergefdrv, line 92: Can you think of a good reason to let a node fall for more than one epoch benind?

matejpavlovic: To avoid false-positives in detecting nodes that fall behind. I expect this parameter to end up being 1, but it might be interesting to experiment with higher values at some point too.

pkg/iss/iss.go
```diff
@@ -605,6 +596,36 @@ func (iss *ISS) applyStableCheckpoint(stableCheckpoint *isspb.StableCheckpoint)
...
+func (iss *ISS) applyPushCheckpoint() (*events.EventList, error) {
+
+	// Schedule the next timer event.
+	// The Repeat function of the Timer is not used here,
+	// because this periodic event cannot be explicitly garbage-collected.
+	eventsOut := events.ListOf(events.TimerDelay(
+		timerModuleName,
+		[]*eventpb.Event{PushCheckpoint()},
+		t.TimeDuration(iss.config.CatchUpTimerPeriod),
+	))
```
sergefdrv, line 608: This potentially delays sending the stable checkpoint up to `CatchUpTimerPeriod`. I think delayed nodes should get assistance as soon as possible, since the more we wait the more delayed they become, and, under high load, by the time they receive the stable checkpoint they may be already too far behind again. 

Moreover, I don't think we can guarantee that the time it takes to finish an epoch is strictly less than `CatchUpTimerPeriod / RetainedEpochs`.

pkg/iss/config.go
sergefdrv: Is it because we artificially chose to create checkpoints at the same pace as we advance through ISS epochs?

matejpavlovic: Not necessarily. Even temporary glitches in a node's performance or network connectivity might make it seem to have fallen behind.

But my main reasoning behind having this parameter is that it costs almost nothing to be able to tune this value instead of having a hard-coded 1. So the very worst case is that it will always be set to 1. And if we do want to try a different value, we have a parameter for it.

sergefdrv: > temporary glitches in a node's performance or network connectivity

... is exactly the case to treat the node as fallen behind, no?

Since the main reason for checkpoints is to perform garbage-collection of protocol history, it doesn't seem particularly useful to keep more of it than up to one stable checkpoint behind.

pkg/iss/iss.go
sergefdrv: The former implementation sent the StableCheckpoint message right away. We could have just added periodic re-transmission of that message. 

Again, under heavy load, it may be too late to try catching up after a few seconds. Namely, if the rest of the system keeps advancing for more than `RetainedEpochs` while waiting for `CatchUpTimerPeriod` before sending StableCheckpoint messages to the delayed node then the delayed node has no chance to catch up.

sergefdrv: Or even without re-transmission, the delayed node would just need to wait until the next stable checkpoint is received in time.

pkg/iss/config.go
sergefdrv: > it costs almost nothing to be able to tune this value

Too many configuration parameters may be confusing for the user.

matejpavlovic: The node with the checkpoint can also experience network connectivity glitches, making it falsely suspect another node to have fallen behind. And as I said, it is very well possible that the parameter will end up being 1 most of the time.

> Too many configuration parameters may be confusing for the user.

I don't think this is a big issue, if the parameter has its default value and the user doesn't ever have to touch it.

Anyway, if you feel strongly about this and want to have a hard-coded 1 instead of the parameter, go ahead and push the change directly to this branch (GH should let you do that), it's fine with me, but pls let's move on in this issue.

pkg/iss/iss.go
matejpavlovic: Fair enough. I still think that even the independent catchup mechanism works in practice and I find it cognitively simpler. But it's true that sending the checkpoint right away helps make the situation you described less likely. Simply periodically re-transmitting the message with the checkpoint could get rather wasteful, should the checkpoint consist of more than just a simple digest. I thus implemented the direct catching up right after the stable checkpoint, but the re-transmission still happens using a periodic internal event upon which the node checks whom to send the checkpoint to.

sergefdrv: Let it be)

pkg/iss/config.go
sergefdrv: Let it be)

pkg/iss/config.go
```diff
@@ -177,6 +190,10 @@ func DefaultConfig(membership []t.NodeID) *Config {
 		LeaderPolicy:                 &SimpleLeaderPolicy{Membership: membership},
 		RequestNAckTimeout:           16,
 		MsgBufCapacity:               32 * 1024 * 1024, // 32 MiB
+		RetainedEpochs:               2,
```
sergefdrv, line 193:
```suggestion
		RetainedEpochs:               1,
```

PR #118, opened 2022-07-04
De-ISS-ize node.go

xosmig:
Remove hardcoded dependencies on ISS from `node.go`, the main entry point of the Mir framework.
Additionally, add the Init event for all modules (previously, it was only sent to a module with id `"iss"`) and add a simple unit test for `node.Run`.


xosmig:
With this PR I hope to kick off the process of de-ISS-ation :)
I think we all agree that Mir should exist beyond ISS.
However, currently, the testing and debugging infrastructure are very ISS-specific and, until this PR, it was impossible to even create a node without the iss module.

matejpavlovic (approved):
👍

PR #119, opened 2022-07-05
Fix creating and removing pre-defined test deployment directory

sergefdrv:
The code in `runIntegrationWithISSConfig` expects the deployment directory under the system temporary directory. Create the test deployment directory there if the test config specifies it explicitly. Moreover, ensure the directory gets removed when the test is over.


matejpavlovic (approved):
👍

PR #122, opened 2022-07-07
Keep resending critical messages in ISS

matejpavlovic:
The only one that was still missing was the checkpoint message.
The rest is either not critical for correctness or already being re-transmitted.

This is just the minimal modification to guarantee correctness of the protocol. Even some non-critical protocol messages might be periodically re-transmitted to avoid view changes in certain cases, but that is future work for another PR.


sergefdrv (approved):
:+1:

PR #126, opened 2022-07-11
Add libp2p

dnkolegov:
This PR adds support for libp2p transport.


pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package libp2p
```
adlrocha, line 7: Maybe add here a `//go:generate go run ./gen/gen.go` directive to re-generate the cbor_gen automatically with `go generate`?

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (t *Transport) Stop() {
+	t.logger.Log(logging.LevelDebug, "Stopping libp2p transport.")
+	defer t.logger.Log(logging.LevelDebug, "libp2p transport stopped.")
+
+	t.outboundStreamsMx.Lock()
+	defer t.outboundStreamsMx.Unlock()
+
+	for id, s := range t.outboundStreams {
```
adlrocha, line 86: I guess this is redundant. IIRC `host.Close()` implicitly closes all the active streams, but this is fine (is a good sanity-check).

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+			t.logger.Log(logging.LevelDebug, fmt.Sprintf("node %s has connected to node %s", t.ownID.Pb(), nodeID.Pb()))
+
+		}(nodeID, nodeAddr)
+	}
+
+	wg.Wait()
+}
+
+func (t *Transport) openStream(ctx context.Context, p peer.ID) (network.Stream, error) {
+	timeout := 500 * time.Millisecond
```
adlrocha, line 151: Out of a curiosity, what is the reason for this timeout?

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+		cancel()
+		if err != nil {
+			t.logger.Log(logging.LevelDebug, "opening stream", "err", err.Error())
+			select {
+			case <-time.After(timeout):
+				continue
+			case <-ctx.Done():
+				return nil, fmt.Errorf("context closed")
+			}
+		}
```
adlrocha, line 164: I don't understand what you are doing here. Why is using the `context.WithTimeout` not enough to timeout if the connection doesn't succeed? Why are you checking the timeout and the ctx.Done(), that is redundant, right? 

Actually, if I am not mistaken, if the timeout ends, `NewStram` will return a context cancel error and that's it.

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (t *Transport) Send(dest types.NodeID, payload *messagepb.Message) error {
+	outBytes, err := proto.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("failed to marshal message: %w", err)
+	}
+
+	t.outboundStreamsMx.Lock()
+	s, ok := t.outboundStreams[dest]
+	t.outboundStreamsMx.Unlock()
```
adlrocha, line 177: If we expect several routines to write at the same time to the stream (i.e. they are able to send messages in parallel) we should defer a bit the release of the lock. Writers are not thread-safe: https://stackoverflow.com/questions/19646717/is-the-go-bytes-buffer-thread-safe

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	return err
+}
+
+func (t *Transport) mirHandler(s network.Stream) {
+	t.logger.Log(logging.LevelDebug, fmt.Sprintf("mir handler for %s started", s.ID()))
+	defer t.logger.Log(logging.LevelDebug, fmt.Sprintf("mir handler for %s stopped", s.ID()))
+
+	defer s.Reset() // nolint
```
adlrocha, line 210: Why Reset instead of Close?

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+		var msg TransportMessage
+		err := msg.UnmarshalCBOR(s)
+		if err != nil {
+			t.logger.Log(logging.LevelError, "failed to read mir transport request", "err", err)
+			return
+		}
+
+		var payload messagepb.Message
+
+		if err := proto.Unmarshal(msg.Payload, &payload); err != nil {
```
adlrocha, line 222: I know it is due to the message format used in Mir and the fact that protobuf can't be sent directly sent through libp2p, but we should maybe consider in the future using exclusively CBOR

adlrocha (approved):
Amazing job, @dnkolegov. Looks good for a first iteration.

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
```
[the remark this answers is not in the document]
dnkolegov: @matejpavlovic what do you think?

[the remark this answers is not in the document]
matejpavlovic: Nope it's not :-)
You must have accidentally copied it along with the content of another file.
You can delete that.

pkg/net/libp2p/libp2ptransport.go
matejpavlovic: Yes, the whole event and message serialization could use some cleaning up. I've created an issue for that: #129

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+
+	wg.Wait()
+}
+
+func (t *Transport) openStream(ctx context.Context, p peer.ID) (network.Stream, error) {
+	timeout := 500 * time.Millisecond
+	for {
+		sctx, cancel := context.WithTimeout(ctx, timeout)
+		s, err := t.host.NewStream(sctx, p, ID)
+		cancel()
```
matejpavlovic, line 152: Why is the context canceled here? What would happen if it wasn't?

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+
+		if err == nil {
+			return s, nil
+		}
+
+		t.logger.Log(logging.LevelError, "opening stream", "err", err.Error())
```
matejpavlovic, line 158: It looks like this should go before `NewStream()`, shouldn't it? At this point, the stream is already opened (or opening failed).

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+			t.logger.Log(logging.LevelDebug, fmt.Sprintf("node %s has connected to node %s", t.ownID.Pb(), nodeID.Pb()))
+
+		}(nodeID, nodeAddr)
+	}
+
+	wg.Wait()
+}
+
+func (t *Transport) openStream(ctx context.Context, p peer.ID) (network.Stream, error) {
+	timeout := 500 * time.Millisecond
```
matejpavlovic, line 148: Can we make this a parameter of the Transport at creation?
Hard-coded literal values deep in the code always give me goosebumps...

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+
+		t.logger.Log(logging.LevelError, "opening stream", "err", err.Error())
+
+		select {
+		case <-time.After(timeout):
+			continue
+		case <-ctx.Done():
+			return nil, fmt.Errorf("context closed")
+		}
```
matejpavlovic, line 165: I don't really understand this block... What would happen if we completely removed it?

pkg/net/libp2p/libp2ptransport.go
matejpavlovic: In this implementation, sending happens sequentially. `addOutboundStream` seems to also access this map, so the lock might be necessary. In general, this is not the most efficient implementation, as sending to different destinations could be done in parallel, but this implementation is simple and will do for the moment.

matejpavlovic (commented):
LGTM, just a few comments / questions. Didn't dive too deeply into the technicalities of libp2p, trust you on that ;)

matejpavlovic:
Oh and one more important thing: don't forget to squash the commits before merging :-D

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -0,0 +1,264 @@
...
+			if err != nil {
+				t.logger.Log(logging.LevelError,
+					fmt.Sprintf("failed to parse addr %v for node %v: %v", maddr, nodeID, err))
+				return
+			}
+
+			t.host.Peerstore().AddAddrs(info.ID, info.Addrs, peerstore.PermanentAddrTTL)
+
+			ctx, cancel := context.WithTimeout(ctx, defaultMaxTimeout)
+			defer cancel()
```
matejpavlovic, line 135: Is there a purpose for canceling the context here? it seems that when this line is reached, nobody looks at the context anyway any more...

matejpavlovic (approved):
👍

PR #17, opened 2022-04-21
Use string as IDs

dnkolegov:
This PR fixes #5.

PR #130, opened 2022-07-12
Remove some misleading comments from the past

xosmig:
A minor clean-up after #17


matejpavlovic (approved):
Thanks, this has been long overdue :-)

PR #136, opened 2022-07-15
WAL interface

sergefdrv:
Define WAL interface and pass it explicitly to `NewNode`. This allow us to replace the hacky way of grabbing the event stored in WAL at initialization:
```go 
n.modules["wal"].(modules.PassiveModule).ApplyEvents(events.ListOf(events.WALLoadAll("wal")))
```


pkg/wal/iface.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/filecoin-project/mir/pkg/events"
+)
+
+// WAL provides access to the write-ahead log.
+//
+// WALLoadAll method loads all events stored in the WAL and returns
+// them as a new EventList.
+type WAL interface {
+	WALLoadAll(ctx context.Context) (*events.EventList, error)
```
matejpavlovic, line 18:
```suggestion
	LoadAll(ctx context.Context) (*events.EventList, error)
```
... and the corresponding implementation and usage

matejpavlovic (approved):
Nice, thanks!

PR #141, opened 2022-07-15
Implement MockPassiveModule

### LINKED ISSUE #138: Mock modules
Standard mocking utils like [gomock](https://github.com/golang/mock) do not really work well with Mir because they only allow setting expectations on method calls while we need to set expectations on the received events and, instead of specifying return values, we should be able to specify output events.

xosmig:
Unlike a simple mock object that could be generated by gomock for
the PassiveModule interface, MockPassiveModule allows setting
expectations on events and not on calls to ApplyEvents.

part of #138


matejpavlovic:
Sorry I won't be able to review this for a week, feel free to merge without my review.

And watch out for "Partly resolves #138", as GitHub doesn't care about "partly", AFAIK. If you merge this, the issue probably will be automatically resolved.

xosmig:
> Sorry I won't be able to review this for a week

No problem :)

> watch out for "Partly resolves https://github.com/filecoin-project/mir/issues/138"

Thanks, replaced with "part of #138".

pkg/modules/mockmodules/internal/impl.go
```diff
@@ -0,0 +1,10 @@
+package internal
+
+import (
+	"github.com/filecoin-project/mir/pkg/events"
+	"github.com/filecoin-project/mir/pkg/pb/eventpb"
+)
+
+type ModuleImpl interface {
+	Event(ev *eventpb.Event) (*events.EventList, error)
```
sergefdrv, line 9: I wonder how well `gomock` matches `events.EventList`. Would you mind to try matching something other than `events.EmptyList()`?

xosmig: Afaik, it doesn't need to match the return value, only the parameters (in this case, an event).

But in case you are still wondering, it does seem to work as expected with some non-empty event lists (see [node_test.go](https://github.com/filecoin-project/mir/blob/main/node_test.go), for example). Internally, it simply uses `reflect.DeepEqual`.

However, it doesn't work well with protobufs in certain cases since they may have some nasty unexported fields.
I'll patch it up a bit in another PR.

PR #143, opened 2022-07-18
De-ISS-ize RequestReceiver

xosmig:
Remove hardcoded dependencies of RequestReceiver on ISS.


matejpavlovic (approved):
Eventually, the request receiver could be made an active module (more idiomatic Mir design). Created #151 for it, not to be addressed in this PR.

PR #144, opened 2022-07-18
Simple mempool module prototype

xosmig:
The prototype expects that new transactions will be delivered to it by means of `NewRequests` events and stores them in an in-memory map. It uses hashes for transaction and batch IDs. 
To support crash-recovery, persistent storage should be added for the transactions.

### DIFF
--- a/pkg/mempool/simplemempool/internal/common/common.go
+++ b/pkg/mempool/simplemempool/internal/common/common.go
@@ -1,0 +1,24 @@
+package common
+
+import (
+	"github.com/filecoin-project/mir/pkg/pb/requestpb"
+	t "github.com/filecoin-project/mir/pkg/types"
+)
+
+// ModuleConfig sets the module ids. All replicas are expected to use identical module configurations.
+type ModuleConfig struct {
+	Self   t.ModuleID // id of this module
+	Hasher t.ModuleID
+	Crypto t.ModuleID
+}
+
+// ModuleParams sets the values for the parameters of an instance of the protocol.
+// All replicas are expected to use identical module parameters.
+type ModuleParams struct {
+	MaxTransactionsInBatch int
+}
+
+// State represents the common state accessible to all parts of the module implementation.
+type State struct {
+	TxByID map[t.TxID]*requestpb.Request
+}
--- a/pkg/mempool/simplemempool/internal/parts/formbatches/formbatches.go
+++ b/pkg/mempool/simplemempool/internal/parts/formbatches/formbatches.go
@@ -1,0 +1,75 @@
+package formbatches
+
+import (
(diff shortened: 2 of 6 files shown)


pkg/mempool/simplemempool/internal/parts/formbatches/formbatches.go
```diff
@@ -0,0 +1,95 @@
...
+	state := &State{
+		State:    commonState,
+		NewTxIDs: nil,
+	}
+
+	dsl.UponNewRequests(m, func(requests []*requestpb.Request) error {
+		txs := make([][]byte, len(requests))
+		for i, req := range requests {
+			var err error
+			txs[i], err = proto.Marshal(req)
```
matejpavlovic, line 37: Protobuf marshalling is not guaranteed to be deterministic. If two nodes receive the same request, they might assign it different IDs in their respective mempools. Not 100% sure this is a correctness issue, but if nothing else, it might be confusing for someone trying to understand the handling of requests.
Custom (deterministic) serialization of a request would be nice here.

xosmig:
depends on #164

pkg/mempool/simplemempool/internal/common/common.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// ModuleParams sets the values for the parameters of an instance of the protocol.
+// All replicas are expected to use identical module parameters.
+type ModuleParams struct {
+	MaxBatchSizeInBytes    int
+	MaxTransactionsInBatch int
+}
+
+// State represents the common state accessible to all parts of the multisig collector implementation.
```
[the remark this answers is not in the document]
xosmig: Fixed

pkg/mempool/simplemempool/internal/parts/formbatches/formbatches.go
xosmig: Fixed

xosmig:
I have no idea why the linter in the CI fails. It works locally on my laptop.
It says that some files are not `gofmt`ed, but `gofmt -l .` outputs no files.

xosmig:
The linter complains about files that were not changed in this PR, so I will just ignore it.

image

matejpavlovic (approved):
Nice!
For the linter, it looks like it has a problem with indented text in comments.
But rather than modifying the comments. But I'd see that rather as a linter problem that is to be fixed later.

matejpavlovic:
Actually, the comments can easily be fixed. Sergey had the same problem and this is the fix: https://github.com/filecoin-project/mir/pull/178/commits/888d9cc064664f0947afd9a9ee586f85bb18ace3

PR #145, opened 2022-07-18
Fix minor issues in dsl wrappers for mempool events

xosmig:


xosmig:
Will merge without a review.

PR #152, opened 2022-07-25
Add Stop method to Node

matejpavlovic:
This adds a cleaner way of stopping the node, so one does not have to cancel the context given to the Run method.

### DIFF
--- a/samples/chat-demo/main.go
+++ b/samples/chat-demo/main.go
@@ -281,7 +281,7 @@ func run() error {
 	if args.Verbose {
 		fmt.Println("Stopping server.")
 	}
-	ctx.Done()
+	node.Stop()
 	wg.Wait()
 
 	return nodeErr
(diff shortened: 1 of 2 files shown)


dnkolegov (commented):
LGTM! But tests do not work. I reran them several times but they failed.

samples/chat-demo/main.go
```diff
@@ -281,7 +281,7 @@ func run() error {
 	if args.Verbose {
 		fmt.Println("Stopping server.")
 	}
-	ctx.Done()
+	node.Stop()
 	wg.Wait()
```
sergefdrv, line 285: Apparently, we don't need this `wg` anymore.

sergefdrv (approved):

samples/chat-demo/main.go
matejpavlovic: Good point. Fixed.

PR #155, opened 2022-07-26
Update Mir network transport

dnkolegov:
This PR makes several changes in network transport to support reconfiguration in Eudico.


pkg/net/libp2p/libp2ptransport.go
```diff
@@ -87,47 +86,50 @@ func (t *Transport) Stop() {
...
 
-func (t *Transport) Connect(ctx context.Context) {
+func (t *Transport) Connect(ctx context.Context, nodes map[types.NodeID]types.NodeAddress) {
 	wg := sync.WaitGroup{}
-	wg.Add(len(t.membership) - 1)
+	if len(nodes) == 0 {
+		t.logger.Log(logging.LevelWarn, "failed to connect: empty map of nodes")
+		return
+	}
+	wg.Add(len(nodes) - 1)
```
matejpavlovic, line 113: This takes a bit of brain gymnastics to read without any comments.
Could you either add a comment saying that`-1` is for the own connection, or even remove the `-1` and add a `wg.Done()` before line 117?
I think that would be easier to read.

matejpavlovic (approved):
Nice!
I like the idea of only using the `Connect` method to update the connections.
Could you also add a way of closing the connections that are *not* part of the new membership?

pkg/net/libp2p/libp2ptransport.go
dnkolegov: I agree. Fixed.

dnkolegov:
> Could you also add a way of closing the connections that are not part of the new membership?

@matejpavlovic I have added a `CloseOldConnections` method.

matejpavlovic (approved):
👍

PR #157, opened 2022-07-27
Fix a timeout in dsl tests

xosmig:


xosmig:
Fixes a flaky test. Will merge without a review.

PR #166, opened 2022-08-02
Make Init event optional in DSL modules

xosmig:
Most modules don't need to do anything on Init and the philosophy
of DSL modules is to avoid mixing protocol logic with boilerplate.
Hence, it makes sense to avoid writing an empty handler for init
for each DSL module and just make processing this event optional.


xosmig:
Will merge without a review.

PR #168, opened 2022-08-02
Add a dsl wrapper for NewRequests events

xosmig:


xosmig:
Will merge without a review

PR #176, opened 2022-08-04
Bcb demo

xosmig:
In the demo, the first node sends a single message to all other nodes using Byzantine Consistent Broadcast.


matejpavlovic (approved):
👍

PR #177, opened 2022-08-05
Benchmarking tools

### LINKED ISSUE #174: Load generator for performance evaluation
Implement an application that submits requests to a Mir cluster at specified rate. It should enable measuring latency and throughput.

sergefdrv:
This PR resolves #174.

### DIFF
--- a/cmd/bench/cmd/node.go
+++ b/cmd/bench/cmd/node.go
@@ -20,7 +20,7 @@ import (
 	"github.com/filecoin-project/mir/pkg/availability/batchdb/fakebatchdb"
 	"github.com/filecoin-project/mir/pkg/availability/multisigcollector"
 	"github.com/filecoin-project/mir/pkg/batchfetcher"
-	mirCrypto "github.com/filecoin-project/mir/pkg/crypto"
+	"github.com/filecoin-project/mir/pkg/deploytest"
 	"github.com/filecoin-project/mir/pkg/iss"
 	"github.com/filecoin-project/mir/pkg/logging"
 	"github.com/filecoin-project/mir/pkg/membership"
@@ -163,17 +163,14 @@ func runNode() error {
 	// retrieves the corresponding transaction batches, and delivers them to the application.
 	batchFetcher := batchfetcher.NewModule(batchfetcher.DefaultModuleConfig())
 
-	// Use dummy crypto module that only produces signatures
-	// consisting of a single zero byte and treats those signatures as valid.
-	// TODO: Adjust once a default crypto implementation is provided by Mir.
-	crypto := mirCrypto.New(&mirCrypto.DummyCrypto{DummySig: []byte{0}})
+	cryptoSystem := deploytest.NewLocalCryptoSystem("pseudo", membership.GetIDs(initialMembership), logger)
 
 	stats := NewStats()
 	interceptor := NewStatInterceptor(stats, "app")
 
 	nodeModules, err := iss.DefaultModules(modules.Modules{
 		"net":          transport,
-		"crypto":       crypto,
+		"crypto":       cryptoSystem.Module(ownID),
 		"iss":          issProtocol,
 		"app":          &App{Logger: logger, ProtocolModule: "iss", Membership: nodeAddrs},
 		"batchfetcher": batchFetcher,
--- a/cmd/bench/cmd/node.go
+++ b/cmd/bench/cmd/node.go
@@ -6,9 +6,12 @@ package cmd
 
 import (
 	"context"
+	"encoding/csv"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -33,6 +36,8 @@ const (
 
 var (
 	transportType string
+	statFileName  string
+	statPeriod    time.Duration
 
 	nodeCmd = &cobra.Command{
 		Use:   "node",
@@ -46,6 +51,8 @@ var (
 func init() {
 	rootCmd.AddCommand(nodeCmd)
 	nodeCmd.Flags().StringVarP(&transportType, "net", "t", "libp2p", "network transport (libp2p|grpc)")
+	nodeCmd.Flags().StringVarP(&statFileName, "statFile", "o", "", "output file for statistics")
+	nodeCmd.Flags().DurationVar(&statPeriod, "statPeriod", time.Second, "statistic record period")
 }
 
 func runNode() error {
@@ -105,6 +112,9 @@ func runNode() error {
 	// TODO: Adjust once a default crypto implementation is provided by Mir.
 	crypto := mirCrypto.New(&mirCrypto.DummyCrypto{DummySig: []byte{0}})
 
+	stats := NewStats()
+	interceptor := NewStatInterceptor(stats)
+
 	nodeModules, err := iss.DefaultModules(modules.Modules{
 		"net":    transport,
 		"crypto": crypto,
@@ -116,7 +126,7 @@ func runNode() error {
 	}
 
 	nodeConfig := &mir.NodeConfig{Logger: logger}
-	node, err := mir.NewNode(t.NodeID(id), nodeConfig, nodeModules, nil, nil)
+	node, err := mir.NewNode(t.NodeID(id), nodeConfig, nodeModules, nil, interceptor)
 	if err != nil {
 		return fmt.Errorf("could not create node: %w", err)
 	}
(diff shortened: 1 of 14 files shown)


cmd/bench/cmd/root.go
```diff
@@ -0,0 +1,39 @@
...
+	id      string
+	nrNodes int
+	verbose bool
+
+	rootCmd = &cobra.Command{
+		Use:   "bench",
+		Short: "Mir benchmarking tool",
+		Long: `Mir benchmarking tool can run a Mir nodes, measuring latency and
+throughput. The tool can also generate and submits requests to the Mir
+cluster at a specified rate.`,
```
matejpavlovic, line 23:
```suggestion
		Long: `Mir benchmarking tool can run Mir nodes, measuring latency and
		throughput. The tool can also generate and submit requests to the Mir
		cluster at a specified rate.`,
```

matejpavlovic (approved):
Nice! 👍

cmd/bench/cmd/root.go
sergefdrv: I don't think we want the extra whitespace in the description on terminal output.

xosmig:
@sergefdrv you'll need to rebase again if you want a green checkmark from the linter.

cmd/bench/cmd/root.go
matejpavlovic: Ok, feel free to not add the whitespace :)

xosmig, line 23:
```suggestion
		Long: "Mir benchmarking tool can run a Mir nodes, measuring latency " +
			"and throughput. The tool can also generate and submit requests " +
			"to the Mir cluster at a specified rate.",
```

Assuming you don't want the output to be word-wrapped.

cmd/bench/cmd/root.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+var (
+	id      string
+	nrNodes int
+	verbose bool
+
+	rootCmd = &cobra.Command{
+		Use:   "bench",
+		Short: "Mir benchmarking tool",
```
xosmig, line 20: Now it looks more like "ISS benchmarking tool".
Please keep in mind that Mir != ISS.

cmd/bench/cmd/node.go
```diff
@@ -0,0 +1,182 @@
...
+	issConfig := iss.DefaultConfig(nodeIDs)
+	issProtocol, err := iss.New(t.NodeID(id), issConfig, logger)
+	if err != nil {
+		return fmt.Errorf("could not instantiate ISS protocol module: %w", err)
+	}
+
+	// Use dummy crypto module that only produces signatures
+	// consisting of a single zero byte and treats those signatures as valid.
+	// TODO: Adjust once a default crypto implementation is provided by Mir.
+	crypto := mirCrypto.New(&mirCrypto.DummyCrypto{DummySig: []byte{0}})
```
xosmig, line 113: What exactly do we want to measure with this benchmark?
If it is the performance of ISS, then crypto (I guess) is an important part and you cannot use dummy crypto.

cmd/bench/cmd/node.go
```diff
@@ -0,0 +1,182 @@
...
+
+	reqReceiverAddrs := make(map[t.NodeID]string)
+	for i := range nodeIDs {
+		reqReceiverAddrs[t.NewNodeIDFromInt(i)] = fmt.Sprintf("127.0.0.1:%d", ReqReceiverBasePort+i)
+	}
+
+	var transport net.Transport
+	nodeAddrs := make(map[t.NodeID]t.NodeAddress)
+	switch strings.ToLower(transportType) {
+	case "grpc":
```
xosmig, line 86: Depending on what we want to measure, local grpc/libp2p transport may not make much sense (there will be no network delays, only marshaling/unmarshaling delays).

xosmig (changes requested):
Would it be possible to do a more general benchmarking library for mir and then just instantiate it here for ISS?

go.mod
```diff
@@ -28,7 +28,9 @@ require github.com/golang/mock v1.6.0
 require (
 	github.com/ipfs/go-cid v0.1.0
 	github.com/multiformats/go-multiaddr v0.5.0
+	github.com/spf13/cobra v1.5.0
```
xosmig, line 31: Demos in `samples` use `kingpin`.
I don't have a preference, but we should probably aim at being consistent.

cmd/bench/cmd/root.go
sergefdrv: It's still not perfectly indented in the source code, though looks a little better

sergefdrv:
> Would it be possible to do a more general benchmarking library for mir and then just instantiate it here for ISS?

I think it would make more sense once we have another atomic broadcast protocol to benchmakr.

cmd/bench/cmd/root.go
sergefdrv: It seems that so far Mir's main protocol implementation is ISS. We can generalize it later.

cmd/bench/cmd/node.go
sergefdrv: @matejpavlovic Do we have a better Crypto module implementation to banchmark with?

cmd/bench/cmd/node.go
sergefdrv: The tool would be enhanced with the ability to run on different machines.

go.mod
sergefdrv: I considered both packages and `cobra` looked more popular. I used `cobra` before and it was easy to deal with.

sergefdrv:
@xosmig since I'm on vacation, could you please merge this PR if you're fine with how I address your comments?

xosmig:
> I think it would make more sense once we have another atomic broadcast protocol to benchmark.

We will probably need to benchmark the availability layer soon (in fact, it could even make sense to do it now, with multisigcollector).

cmd/bench/cmd/root.go
xosmig: Please see the discussion section [here](https://github.com/protocol/ConsensusLab/issues/9#issuecomment-1208245086)

xosmig:
> @xosmig since I'm on vacation

Enjoy your vacation! :)

> could you please merge this PR if you're fine with how I address your comments?

let's return to discussing this PR after you come back from your vacation.

cmd/bench/cmd/node.go
matejpavlovic: Yes, there is a ["Pseudo" crypto implementation](https://github.com/filecoin-project/mir/blob/main/pkg/crypto/pseudo.go#L29) for this purpose. It actually performs all the cryptographic operations, just the keys are generated deterministically, so it's easier to set up. You just need to give it the node IDs on instantiation. [Here's an example](https://github.com/filecoin-project/mir/blob/main/pkg/iss/iss_test.go#L371) of how it's used.

matejpavlovic:
Hi! I think this PR is a pretty good start for the benchmarking tooling. Of course, a lot of things can be extended and generalized. When we have some cycles to actually use it before Sergey comes back from vacation, we can see whether that's reason enough to merge it before generalization (or even generalize it ourselves), or continue improving it as a PR and only merge when more general.

cmd/bench/cmd/root.go
sergefdrv: Indeed, this tool currently supports benchmarking ISS instantiations only, but it could be extended in future, e.g. we could add CLI options or some kind of flexible configuration.

cmd/bench/cmd/node.go
sergefdrv: It seems like now we can do that with the membership configuration file.

matejpavlovic: Btw example membership config files are already in the chat demo. I guess I should have created an example directly within the bench folder...

cmd/bench/cmd/root.go
matejpavlovic: It looks like another layer of abstraction / encapsulation could be useful here that would group related modules together to something we could call "system" or "application" or something like that. I actually created an issue for that yesterday (#215).

We could then create separate benchmarks specific to those module compositions, potentially (re-)using some general benchmarking package (e.g. a customizable interceptor with a configurable event selector function and event processor function etc.). Benchmarking "pure Mir" is anyway only meaningful to obtain some very low-level microbenchmarks (like events processed per second or so). Not saying that it wouldn't be useful, but the most practically meaningful benchmarks inherently need to be model- (or model-composition-) specific.

So for now, we could 1) choose a name for the TOB / SMR implementation we are building now (which is the first and so far only actual application of Mir) 2) rename this benchmark accordingly. WDYT?

sergefdrv:
I will update the PR to use the new “SMR system” abstraction.

xosmig (approved):

PR #182, opened 2022-08-10
Fix transaction counting bug in simple mempool

matejpavlovic:
The current implementation didn't properly count the number of transactions added to the batch.
When there was only one transaction in the mempool, the counter `i` would always keep the value 0 and `newTxIDs` was not trimmed properly.


xosmig (commented):
Thanks! I'll try to add a test for this case later today.

xosmig (approved):

PR #184, opened 2022-08-11
Update all dependencies

xosmig:
Previously, the project wouldn't build with the latest release of
Go at the time (1.19). The reason for it is that the version of
the quic-go package was not compatible with go 1.19.
Unfortunately, older versions of other packages were not
compatible with the latest version of quic-go, so I just updated
all of them.

PR #185, opened 2022-08-11
Fix the latest version of go in the CI config

xosmig:


xosmig:
Depends on #184

dnkolegov (commented):
> No description provided.
What is the goal of this PR?

I would like to remind (and that is not obvious) that at present Mir must use Go modules and features compatible with Eudico.

xosmig:
> What is the goal of this PR?

To "Fix the latest version of go in the CI config" so that it actually uses the latest version of Go.
However, as discussed in #184, apparently, we cannot support Go 1.19 yet.

PR #188, opened 2022-08-15
Trace logging

xosmig:


pkg/logging/logger.go
```diff
@@ -64,6 +72,10 @@ func (l consoleLogger) Log(level LogLevel, text string, args ...interface{}) {
 	fmt.Printf("\n")
 }
 
+func (l consoleLogger) Level() LogLevel {
+	return l.level
+}
+
```
matejpavlovic, line 78: I'm not a big fan of exposing the internal settings of the logger. Somehow it goes against my view of how a logger is used naturally - the application simply announces the importance levels of its log entries and it's only the logger that decides what to do with them.

I see how it can be useful in the application for avoiding to even call the logger, but I think that it's most likely premature optimization (I haven't seen such an optimization make a real performance difference yet). I do see the benefit (enabling low-level optimizations), I just think it's not worth paying the price of breaking the clean logger interface. If we ever see that such an optimization has a non-negligible performance impact, we might come back to this and even use it if we don't find a better solution, but I'd see it as a last-resort option.

workers.go
```diff
@@ -71,6 +72,15 @@ func (n *Node) processModuleEvents(
 	// This is only for debugging / diagnostic purposes.
 	n.interceptEvents(plainEvents)
 
+	// In Trace mode, log all events.
+	if n.Config.Logger.Level() <= logging.LevelTrace {
```
matejpavlovic, line 76: See other comment.

In this particular case, an alternative would be to add tracing to the interceptor or use a "`ConsoleInterceptor`" (in case there is no interceptor) if we want to trace the events. The choice of interceptor could be driven by the same command-line argument that already drives the "verbousness" of the logger.

matejpavlovic (commented):
Very nice. I like most of the improvements, except exposing the level (see comments).

pkg/logging/logger.go
xosmig: > for avoiding to even call the logger

It's not just "calling the logger", sometimes it's doing some (expensive) loops as is the case in this PR.
The whole point of trace logging is to log as much as possible, without worrying about polluting the logs or performance. Without this optimization, we would inevitably have to worry about performance.

> exposing the internal settings of the logger

Tbh, I don't quite see a problem of exposing the `Level()` method as it seems to have a well-defined semantic.
Although, maybe it would make sense to rename the method to `MinLevel()` to make it a bit less ambiguous.

xosmig: I guess a slightly less intrusive option would be a method `TraceEnabled() bool`

matejpavlovic: Yes I do see the point, I'm just asking myself whether expensive loops printing trace output really should be contained in the code and how useful they are (in case of the current PR, the same functionality can be achieved using the event interceptor and placing a simple trace log output within its inner loop).

In case of `TraceEnabled()` vs `MinLevel()`, if we do expose the internal setting of the logger, I think that `MinLevel()` is better. I just meant that this exposition departs from a clean and simple logger interface.

That said, I don't think it's a disaster to have it exposed, and I'm fine with the update. So consider this comment resolved and go ahead with merging :-)

matejpavlovic (approved):

PR #189, opened 2022-08-16
Add the interface and fake implementation for local batch storage

xosmig:


matejpavlovic (approved):
Nice!!
I needed to get used to the idea to have two different handlers for the same event (and convincing myself they don't interfere), but it makes sense. Well done.

PR #190, opened 2022-08-16
Improve libp2p transport

dnkolegov:


pkg/net/libp2p/libp2ptransport.go
```diff
@@ -173,37 +176,44 @@ func (t *Transport) connectToNode(ctx context.Context, addr multiaddr.Multiaddr)
...
 func (t *Transport) openStream(ctx context.Context, p peer.ID) (network.Stream, error) {
-	for {
-		sctx, cancel := context.WithTimeout(ctx, defaultMaxTimeout)
-		s, err := t.host.NewStream(sctx, p, ID)
+	var streamErr error
+	for i := 0; i < retryAttempts; i++ {
+		sctx, cancel := context.WithTimeout(ctx, maxConnectingTimeout)
+
+		s, streamErr := t.host.NewStream(sctx, p, ProtocolID)
 		cancel()
```
matejpavlovic, line 191: Do we need to cancel the context here?
If not, it's a bit confusing to still cancel it.

pkg/net/libp2p/libp2ptransport.go
```diff
@@ -173,37 +176,44 @@ func (t *Transport) connectToNode(ctx context.Context, addr multiaddr.Multiaddr)
...
+			t.logger.Log(
+				logging.LevelInfo, fmt.Sprintf("failed to open stream to %s, retry in %d sec", p, retryTimeout))
+		}
 
+		delay := time.NewTimer(retryTimeout)
 		select {
 		case <-delay.C:
 			continue
 		case <-ctx.Done():
 			if !delay.Stop() {
```
matejpavlovic, line 210: Similar to previous one.
It seems we don't need to stop the timer here, since we're returning anyway, so it would seem much cleaner to me to ignore the timer here.

matejpavlovic (approved):
👍

pkg/net/libp2p/libp2ptransport.go
dnkolegov: Yes we do.

```
WithTimeout returns WithDeadline(parent, time.Now().Add(timeout)).

Canceling this context releases resources associated with it,
so code should call cancel as soon as the operations running in this Context complete
```

pkg/net/libp2p/libp2ptransport.go
dnkolegov: This is a best practice to prevent memory leak - https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/

pkg/net/libp2p/libp2ptransport.go
matejpavlovic: Ok I said nothing :)

pkg/net/libp2p/libp2ptransport.go
matejpavlovic: Fair enough. I still wouldn't really call it a memory leak in this case, since the memory always reclaimed after the timer fires, it just might not happen immediately. But if this is best practice, let's keep it.

PR #191, opened 2022-08-17
Add hierarchical module IDs

matejpavlovic:
In preparation for submodules in Mir this commit introduces structured
module IDs. They are backwards-compatible with the old unstructured
ones, which are considered to be identifiers of top-level modules.

### DIFF
--- a/pkg/types/moduleid.go
+++ b/pkg/types/moduleid.go
@@ -1,0 +1,40 @@
+package types
+
+import "strings"
+
+const Separator = "/"
+
+// TODO: Mention in the documentation that the Separator is a special sequence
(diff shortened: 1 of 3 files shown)


xosmig (approved):
Thanks! 

> `// It corresponds to a path in the module hierarchy.`

In fact, it doesn't even have to be a path in the module hierarchy.
From the Mir core point of view, only the part of the module id until the first separator matters, the rest is just ignored by Mir.
It is up to the module implementations to decide what they do with the ignored part.

matejpavlovic:
Thanks, good points, I still extended the comments, so the note doesn't get lost in an old pull request nobody looks at any more.

PR #193, opened 2022-08-18
Remove new epoch notifications from chat

### LINKED ISSUE #187: Chat demo prints a bunch of warnings to the console
The app prints a bunch of network-related warnings (about connections being refused) in non-verbose mode even when everything works as intended.
I believe, warnings should only be printed when something unexpected happens.

@dnkolegov could you take a look? We probably need to modify the networking module so that it only prints warnings if it cannot establish the connection after some timeout or some unexpected error happens (i.e., anything other than "connection refused" I guess).

matejpavlovic:
Resolves #187


matejpavlovic:
Microscopic change, merging without review.

PR #194, opened 2022-08-19
Generate protobufs first

matejpavlovic:
If the generated protobufs are broken (e.g. during a git rebase), running `make generate` used to return an error, because it tried to compile them first with Go. Now the `./protos` are generated first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ lint:
 
 .PHONY: generate
 generate:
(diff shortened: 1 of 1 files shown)


xosmig (approved):
I wonder if there is a `go generate`-native solution to this problem. Mixing `go generate` and makefiles looks a bit weird as the goal of go generate is to replace `make`.

That said, the solution seems to work, so LGTM :)

PR #199, opened 2022-08-25
Add default handler registration to DSL module

matejpavlovic:
This is useful for modules that perform processing only on some events.
E.g., the default action could be forwarding to another module.

### DIFF
--- a/pkg/dsl/dslmodule.go
+++ b/pkg/dsl/dslmodule.go
@@ -13,10 +13,11 @@ import (
 )
 
 type dslModuleImpl struct {
-	moduleID          t.ModuleID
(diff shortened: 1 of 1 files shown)


xosmig (approved):
This looks like a useful feature, thank you!

xosmig:
This may go well with this change I was thinking about but didn't add to the main branch yet: https://github.com/xosmig/mir/pull/2

xosmig:
Right now, the semantics of "not matching an event" is somewhat ambiguous because, in practice, we have a hierarchy of event types in Mir, but the core of DSL is only aware of the top level.

PR #201, opened 2022-08-25
Batch fetcher

matejpavlovic:
Module for transforming availability certificates into batches of requests


xosmig:
Doesn't the availability layer already provide the API for batch reconstruction (event `RequestTransactions`)?

pkg/batchfetcher/batchfetcher.go
```diff
@@ -0,0 +1,131 @@
...
+		output.Enqueue(&outputItem{
+			event: events.NewEpoch(mc.Destination, t.EpochNr(newEpoch.EpochNr)),
+		})
+		output.Flush(m)
+		return nil
+	})
+
+	// The DeliverCert handler requests the transactions referenced by the received availability certificate
+	// from the availability layer.
+	// TODO: Verify Certificate? Here or elsewhere?
```
xosmig, line 50: I think it is ok to assume that the certificate is valid here (i.e., to push the responsibility of verifying them to the module that sends the `DeliverCert` events).

pkg/batchfetcher/batchfetcher.go
```diff
@@ -0,0 +1,131 @@
...
+	dsl.UponEvent[*eventpb.Event_DeliverCert](m, func(cert *eventpb.DeliverCert) error {
+		// Skip padding certificates. DeliverCert events with nil certificates are considered noops.
+		if cert.Cert.Type == nil {
+			return nil
+		}
+
+		switch c := cert.Cert.Type.(type) {
```
xosmig, line 57: Is this switch-case necessary?
Can't this module be agnostic to the availability layer implementation?

pkg/batchfetcher/config.go
```diff
@@ -0,0 +1,18 @@
...
+// ModuleConfig determines the IDs of the modules the batch fetcher interacts with.
+type ModuleConfig struct {
+	Self         t.ModuleID // Own ID.
+	Availability t.ModuleID // ID of the factory module containing the availability modules.
+	Destination  t.ModuleID // ID of the module to deliver the produced event stream to (usually the application).
+}
+
+func DefaultModuleConfig() *ModuleConfig {
+	return &ModuleConfig{
+		Self:         "batchFetcher",
```
xosmig, line 14: Do we want to use snakeCase for multi-word module IDs? Another option would be to use the go package naming convention (withnoclearseparators).

pkg/batchfetcher/batchfetcher.go
```diff
@@ -0,0 +1,131 @@
...
+		case *availabilitypb.Cert_Msc:
+
+			// TODO: Check whether this makes any sense.
+			if len(c.Msc.BatchId) == 0 {
```
xosmig, line 61: Why is this check important?
We probably should introduce a nicer way to check whether a batch is empty.
I am not actually sure whether this way actually works (depends on the `hasher` implementation).

xosmig (approved):
Looks good! Left a few minor comments (mostly cosmetic).

pkg/batchfetcher/batchfetcher.go
```diff
@@ -0,0 +1,131 @@
...
+			// This is necessary to preserve the order of incoming and outgoing events.
+			item := outputItem{event: nil}
+			output.Enqueue(&item)
+
+			// Request transactions from the availability layer.
+			availabilitydsl.RequestTransactions(
+				m,
+				mc.Availability.Then(t.ModuleID(fmt.Sprintf("%v", epochNr))),
+				cert.Cert,
+				&item,
```
xosmig, line 77: I usually prefer having a separate struct for the context to be more explicit and avoid accidentally using the same context type in two places.

pkg/batchfetcher/batchfetcher.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/filecoin-project/mir/pkg/pb/availabilitypb"
+	"github.com/filecoin-project/mir/pkg/pb/eventpb"
+	"github.com/filecoin-project/mir/pkg/pb/requestpb"
+	t "github.com/filecoin-project/mir/pkg/types"
+)
+
+// NewModule returns a new batch fetcher module.
+// The batch receives events output by the ordering protocol (e.g. ISS)
+// and relays them to the application in the same order.
+// It replaces the DeliverCert events from the input stream by the corresponding ProvideTransactions
```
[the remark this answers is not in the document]
xosmig: Also, `ProvideTransactions` has an `Origin` field, which would be meaningless for anyone but the module that sent the original request.

pkg/batchfetcher/batchfetcher.go
```diff
@@ -0,0 +1,131 @@
...
+	// Map of delivered requests that is used to filter duplicates.
+	// TODO: Implement compaction (client watermarks) so that this map does not grow indefinitely.
+	delivered := make(map[t.ClientID]map[t.ReqNo]struct{})
+
+	// Queue of output events. It is required for buffering events being relayed
+	// in case a DeliverCert event that has arrived before has not yet been transformed to a ProvideTransactions event.
+	var output outputQueue
+
+	// The NewEpoch handler updates the current epoch number and forwards the event to the output.
+	dsl.UponEvent[*eventpb.Event_NewEpoch](m, func(newEpoch *eventpb.NewEpoch) error {
```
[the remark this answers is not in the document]
matejpavlovic: I see. I think it might be a matter of personal preference. Using a wrapper definitely makes the final code look nicer, but I think that even using `dsl.UponEvent` is simple and readable enough, especially in such a simple case. Since the "idiomatic" way involves creating an extra file (or even a package?), personally I think that `dsl.UponEvent` is easier. Definitely for someone who is just starting to use the DSL modules.

If those wrappers were generated (I assume they will be after you merge your code generation update), that would be ideal. In a way, since the `NewEpoch` event is a "legacy" one, i.e., is not associated to a module, there would also be a question of where to put the wrapper...

Long story short, I'd prefer using `dsl.UponEvent` here, and only use wrappers when they are generated and only for events that have a clear association with a module.

pkg/batchfetcher/batchfetcher.go
matejpavlovic: Agree. Adapted comment.

pkg/batchfetcher/batchfetcher.go
matejpavlovic: Good point!
Kicked out the switch.

pkg/batchfetcher/config.go
matejpavlovic: No strong feelings about this. gopackagenamingisfinewithme :)

pkg/batchfetcher/batchfetcher.go
matejpavlovic: Yes, you're right, this is crap, I removed it.

pkg/batchfetcher/batchfetcher.go
matejpavlovic: Yes, I see your point. It does take away some of the beauty of how the context store is used under the hood, but making things explicit usually tends to be a good idea.
Wrapped it in a separate struct.

xosmig: Thanks! Yes, this is not ideal, I agree. I spent a good amount of time trying to figure out a nicer way to do this, but it tends to quickly get a bit too complicated internally (which is not necessarily a bad thing, but I think for now it makes sense to keep things relatively low-level).

pkg/batchfetcher/batchfetcher.go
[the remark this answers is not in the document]
xosmig: > I think it might be a matter of personal preference

The issue here is that sometimes we need to do some pre-processing of events (e.g., replacing the origin with the context). Hence, some events do require wrappers. Then I decided to add wrappers for all events for consistency. The other reason was to convert from protobuf types to proper go types.

> there would also be a question of where to put the wrapper

Such wrappers currently live in `pkg/dsl/events`.

In any case, this discussion is about to become irrelevant after I merge the code generation, so let's just go ahead without the wrappers in this PR. I'll mark the conversation as resolved.

pkg/batchfetcher/config.go
xosmig: No strong feelings on my side either :)
`batchdb` will probably need to be changed to `batchDB` if we adopt the snake case. Other module IDs were single words so far, as far as I can see.

pkg/batchfetcher/batchfetcher.go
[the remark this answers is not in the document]
matejpavlovic: Yes it's better indeed, I was just too lazy to do it.
Fixed.

pkg/batchfetcher/batchfetcher.go
```diff
@@ -0,0 +1,101 @@
...
+// It keeps track of the current epoch (by observing the relayed NewEpoch events)
+// and automatically requests the transactions from the correct instance of the availability module.
+func NewModule(mc *ModuleConfig) modules.Module {
+	m := dsl.NewModule(mc.Self)
+
+	// The current epoch number, as announced by the ordering protocol.
+	epochNr := t.EpochNr(0)
+
+	// Queue of output events. It is required for buffering events being relayed
+	// in case a DeliverCert event that has arrived before has not yet been transformed to a ProvideTransactions event.
```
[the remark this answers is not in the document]
matejpavlovic: Fixed.

pkg/batchfetcher/batchfetcher.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/filecoin-project/mir/pkg/events"
+	"github.com/filecoin-project/mir/pkg/modules"
+	"github.com/filecoin-project/mir/pkg/pb/availabilitypb"
+	"github.com/filecoin-project/mir/pkg/pb/eventpb"
+	"github.com/filecoin-project/mir/pkg/pb/requestpb"
+	t "github.com/filecoin-project/mir/pkg/types"
+)
+
+// NewModule returns a new batch fetcher module.
+// The batch receives events output by the ordering protocol (e.g. ISS)
```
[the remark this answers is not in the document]
matejpavlovic: Fixed.

pkg/batchfetcher/config.go
matejpavlovic: Fixed.

PR #205, opened 2022-08-26
Encapsulate multisig collector factory in a func

matejpavlovic:

### DIFF
--- a/pkg/availability/multisigcollector/multisigcollector.go
+++ b/pkg/availability/multisigcollector/multisigcollector.go
@@ -8,8 +8,12 @@ import (
 	"github.com/filecoin-project/mir/pkg/availability/multisigcollector/internal/parts/certcreation"
 	"github.com/filecoin-project/mir/pkg/availability/multisigcollector/internal/parts/certverification"
 	"github.com/filecoin-project/mir/pkg/dsl"
+	"github.com/filecoin-project/mir/pkg/factorymodule"
+	"github.com/filecoin-project/mir/pkg/logging"
 	"github.com/filecoin-project/mir/pkg/modules"
+	"github.com/filecoin-project/mir/pkg/pb/factorymodulepb"
 	t "github.com/filecoin-project/mir/pkg/types"
+	"github.com/filecoin-project/mir/pkg/util/maputil"
 )
 
 // ModuleConfig sets the module ids. All replicas are expected to use identical module configurations.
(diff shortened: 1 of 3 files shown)


matejpavlovic:
@dnkolegov if you adopt this approach in your Eudico integration, it will make the code much less ugly.
(Also resolves this: https://github.com/filecoin-project/eudico/pull/234#discussion_r955140446 )

dnkolegov:
@matejpavlovic Will do after merge

pkg/availability/multisigcollector/multisigcollector.go
```diff
@@ -48,3 +52,40 @@ func NewModule(mc *ModuleConfig, params *ModuleParams, nodeID t.NodeID) (modules
...
+
+				// Crate a copy of basic module config with an adapted ID for the submodule.
+				submc := *mc
+				submc.Self = mscID
+
+				// Create a new instance of the multisig collector.
+				multisigCollector, err := NewModule(
+					&submc,
+					&ModuleParams{
+						InstanceUID: []byte(mscID),
```
xosmig, line 77: The point of `InstanceUID` is to prevent signature replay attacks between instances of the protocol. It probably makes sense for the protocol using the multisig collector to also employ an instance UID and to use the concatenation of the "parent" `InstanceUID` with `mscID` as the `InstanceUID` for multisig collector.

However, it should probably be addressed in a separate PR. Could you at least put a TODO for it so that it doesn't get lost?

xosmig, line 77: The point of `InstanceUID` is to prevent signature replay attacks between instances of the protocol. It probably makes sense for the protocol using the multisig collector to also employ an instance UID and to use the concatenation of the "parent" `InstanceUID` with `mscID` as the `InstanceUID` for multisig collector.

However, it should probably be addressed in a separate PR. Could you at least put a TODO for it so that it doesn't get lost?

pkg/availability/multisigcollector/multisigcollector.go
```diff
@@ -48,3 +52,40 @@ func NewModule(mc *ModuleConfig, params *ModuleParams, nodeID t.NodeID) (modules
...
+						AllNodes:    mscNodeIDs,
+						F:           (len(mscNodeIDs) - 1) / 2,
```
xosmig, line 79: Works for now, but a bit weird :)
It is likely higher than what other modules will use (likely, (N-1)/3 or lower)

xosmig (approved):
Looks good, thank you!
I left a couple of minor comments, but they can be addressed in later PRs.

pkg/availability/multisigcollector/multisigcollector.go
matejpavlovic: ACK, added TODO and created issue.

pkg/availability/multisigcollector/multisigcollector.go
matejpavlovic: I copied that from your demo :-)
But if it works for the MSC, let's keep it, it's a MSC-specific configuration parameter anyway.

xosmig: > I copied that from your demo :-)
It makes sense for the demo because there are no other modules :)

> But if it works for the MSC, let's keep it
It's just that it is a bit wasteful (the certificate will contain roughly 50% more signatures), but yes, not a big deal.
Maybe leave a TODO though?

Feel free to mark this conversation as resolved with or without TODO :)

matejpavlovic: Yes I see. If other modules have a lower fault tolerance, there is no point waiting for too many signatures here.
I'll add a TODO and merge.

PR #213, opened 2022-08-30
Improve recorder io

dnkolegov:
This PR adds a mechanism to open a file on recorder creation and close the recorder file via Stop function.


matejpavlovic (approved):
Nice, makes the code using the recorder simpler and cleaner 👍

PR #219, opened 2022-09-09
Implement threshold cryptography

abread:
This PR introduces `threshcrypto` - a new module for working with threshold signatures in Mir, and an implementation using BLS12-381 threshold signatures (using the [same underlying library as drand](https://github.com/drand/kyber)).

There was [some discussion on Slack](https://filecoinproject.slack.com/archives/C03C77HN3AS/p1660755716337459) on how to go about this.

It's based on the existing `crypto` module: the dummy and pseudo implementations were kept for testing purposes, but I did not reimplement the `DefaultImpl` as no API consumers exist *yet* and there's no "real" implementations besides TBLS to guide its design.


pkg/threshcrypto/api.go
```diff
@@ -0,0 +1,35 @@
...
+	// Returns the signature (and a nil error) on success, and a non-nil error otherwise.
+	SignShare(data [][]byte) ([]byte, error)
+
+	// VerifyShare verifies that a signature share is valid for the given data.
+	// Returns nil on success (i.e., if the given signature share is valid) and a non-nil error otherwise.
+	VerifyShare(data [][]byte, signatureShare []byte) error
+
+	// Recover constructs a full signature from signature shares over data.
+	// The given array may contain invalid signature shares: as long as at least T distinct signature
+	// shares are valid, Recover will succeed, albeit less efficiently.
```
xosmig, line 27: I have mixed feelings regarding this property.
The implementation you are using does indeed verify the partial signatures in `Recover`.
However, it seems kind of redundant in my opinion as the shares can (and, probably, should) be verified before invoking `Recover`.

Maybe let's make it a property of the particular implementation and not of the abstraction (i.e., move it to the comment for `TBLSInst.Recover`)?

xosmig (commented):
Thanks a lot for the PR! I left a few minor cosmetic comments and a couple of comments w.r.t. to the interface.

pkg/threshcrypto/api.go
xosmig: In fact, it could be a serious performance overhead as verifying signature shares (i.e., doing pairings) is the most computationally heavy part. Not only this API forces the protocol to verify the shares multiple times (once before `Recover` and once inside), but it also prohibits parallelizing the verification.

I wonder why they did it this way.

abread: I agree, it's a good idea to force the optimizable design (verify shares, then recover) right from the start, even if some implementations are sub-optimal.

The implementation of `Recover` for tbls only uses public kyber APIs and is simple, so I think we can land the optimized implementation (without the double verification) here.

pkg/threshcrypto/api.go
```diff
@@ -0,0 +1,35 @@
...
+	// SignShare signs the provided data and returns the resulting signature share.
+	// The data to be signed is the concatenation of all the passed byte slices.
+	// A signature share produced by SignShare is verifiable using VerifyShare.
+	// After obtaining signature shares from T group members, the full signature can be constructed with Recover.
+	// Returns the signature (and a nil error) on success, and a non-nil error otherwise.
+	SignShare(data [][]byte) ([]byte, error)
+
+	// VerifyShare verifies that a signature share is valid for the given data.
+	// Returns nil on success (i.e., if the given signature share is valid) and a non-nil error otherwise.
+	VerifyShare(data [][]byte, signatureShare []byte) error
```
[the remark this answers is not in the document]
abread: For BLS it doesn't matter a lot: signatures (and sig shares) are unique and deterministic, so I think it's equivalent to wait for `t` distinct valid shares before calling `Recover`.

That said, other schemes may benefit from this, and some protocols may want to classify a mismatch as byzantine behavior, so I agree that we should include it.

abread:
@xosmig Thank you for the thorough review! I believe I have integrated all the suggestions

pkg/threshcrypto/tbls_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+		_, err = keys2[i].Recover(data, sigShares)
+		assert.NoError(t, err)
+	}
+}
+
+func keygen(T, N int, seed int64) ([]*TBLSInst, error) {
+	members := make([]types.NodeID, N)
+	for i := range members {
+		members[i] = types.NodeID(fmt.Sprint(i))
```
matejpavlovic, line 124: It would be cleaner to use `types.NodeIDFromInt` here.

abread:
Thank you so much, Matej!
I've updated the PR with your suggestions (and preallocated an array as suggested by the linter)

go.mod
```diff
@@ -26,6 +26,8 @@ require (
 require github.com/golang/mock v1.6.0
 
 require (
+	github.com/drand/kyber v1.1.15
+	github.com/drand/kyber-bls12381 v0.2.3
```
dnkolegov, line 30: Is it safe to use this library? Why not use blst bindings?

xosmig (approved):
Thanks a lot!
**UPD:** I added a few more comments. Most are trivial, but the question about the minimum threshold is important.

pkg/threshcrypto/tbls_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+		_, err = keys[i].Recover(data, sigShares2)
+		assert.NoError(t, err)
+
+		_, err = keys2[i].Recover(data, sigShares)
+		assert.NoError(t, err)
+	}
+}
+
+func keygen(T, N int, seed int64) ([]*TBLSInst, error) {
```
xosmig, line 120: seed is unused

pkg/threshcrypto/tbls.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+// TBLS12381Keygen constructs a set TBLSInst for a given set of member nodes and threshold T
+// using the BLS12-381 pairing, with signatures being points on curve G1,
+// and keys points on curve G2.
+func TBLS12381Keygen(T int, members []t.NodeID, randSource cipher.Stream) ([]*TBLSInst, error) {
+	N := len(members)
+	if !(T >= (N+1)/2 && N > 0) {
+		return nil, fmt.Errorf("TBLS requires t >= (n+1)/2 and n > 0, where n is the number of group members")
+	}
```
xosmig, line 52: Could you give a reference for this? I believe many applications need (t+1, 3t+1) threshold signatures.

pkg/threshcrypto/tbls.go
```diff
@@ -0,0 +1,330 @@
...
+			t:         T,
+			members:   members,
+		}
+	}
+
+	return instances, nil
+}
+
+// Writes the properties of a TBLSInst to an io.Writer
+// Can be read with TBLSInst.UnmarshalFrom
```
xosmig, line 81:
```suggestion
// MarshalTo writes the properties of a TBLSInst to an io.Writer.
// Can be read with TBLSInst.UnmarshalFrom.
```

pkg/threshcrypto/tbls.go
```diff
@@ -0,0 +1,330 @@
...
+
+	inst.privShare.V = keyGroup.Scalar()
+	if err := unmarshalKyber(inst.privShare.V); err != nil {
+		return read, err
+	}
+
+	return read, nil
+}
+
+// constructs signature share for message
```
xosmig, line 257:
```suggestion
// SignShare constructs a signature share for the message.
```

pkg/threshcrypto/tbls.go
```diff
@@ -0,0 +1,330 @@
...
+func (inst *TBLSInst) SignShare(msg [][]byte) ([]byte, error) {
+	return inst.scheme.Sign(inst.privShare, digest(msg))
+}
+
+// verifies that a signature share is for a given message from a given node
```
xosmig, line 262:
```suggestion
// VerifyShare verifies that a signature share is for a given message from a given node.
```

pkg/threshcrypto/tbls.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	if idx != slices.Index(inst.members, nodeID) {
+		return fmt.Errorf("signature share belongs to another node")
+	}
+
+	return inst.scheme.VerifyPartial(inst.public, digest(msg), sigShare)
+}
+
+// verifies that a (full) signature is valid for a given message
```
xosmig, line 276:
```suggestion
// VerifyFull verifies that a (full) signature is valid for a given message.
```

pkg/threshcrypto/tbls.go
```diff
@@ -0,0 +1,330 @@
...
+func (inst *TBLSInst) VerifyFull(msg [][]byte, sigFull []byte) error {
+	return inst.scheme.VerifyRecovered(inst.public.Commit(), digest(msg), sigFull)
+}
+
+// recovers a full signature from a set of (previously validated) shares, that are known to be from
+// distinct nodes
```
xosmig, line 282:
```suggestion
// Recover recovers a full signature from a set of (previously validated) shares, that are known to be from
// distinct nodes.
```

pkg/threshcrypto/tbls.go
```diff
@@ -0,0 +1,330 @@
...
+	if err := marshalKyber(inst.privShare.V); err != nil {
+		return written, err
+	}
+
+	return written, nil
+}
+
+// Sets the properties of a TBLSInst from an io.Reader
+// The property stream can be created from TBLSInst.MarshalTo
+// NOTE: Currently assumes the underlying scheme is the same as in TBLS12381Keygen()
```
xosmig, line 157:
```suggestion
// UnmarshalFrom sets the properties of a TBLSInst from an io.Reader.
// The property stream can be created from TBLSInst.MarshalTo.
// NOTE: Currently assumes the underlying scheme is the same as in TBLS12381Keygen().
```

pkg/threshcrypto/tbls.go
abread: Ah oops, I had a "TODO" comment precisely to remember to review this check carefully, and must have accidentally deleted it at some point

abread: Ok, I reviewed the [literature](https://link.springer.com/content/pdf/10.1007/3-540-36288-6_3.pdf), and comitted a correct version, that supports the (t+1, 3t+1) use case.

It's hard (impossible?) to dissociate the check from the number of malicious nodes.

xosmig: I would suggest removing this check altogether as it is rather trivial and concerns rather the protocol using the threshold signatures than the threshold signatures themselves.

I.e., you can have a (2, n)-threshold signature scheme, why not. It's just that, in the classical BFT model, you are unlikely to find such a signature useful, but it is not a concern of the threshold signature module.

xosmig: Or am I missing something?
Where does `nByz < (N+1)/2` come from?

xosmig: My speculation is that this is the requirement not for the threshold signature scheme itself, but for the interactive signing protocol. However, I need to check.

abread:
I've started playing around with this, and ended up implementing some DSL utilities akin to the availability module.

They don't require much thought to implement so I went ahead and included them here

pkg/threshcrypto/tbls.go
xosmig: Sorry, for taking so much time. I believe that the `t < n/2` part in the paper comes from the way the threshold signature scheme is used (in particular, the DKG part and the protocol for signing a message), which is kind of irrelevant.
So, I stick to my suggestion to remove this check completely.

xosmig:
@abread sorry that you had to do it several times already, but could you please merge the latest `main` branch into your branch?
Also, would it be OK if we use the "squash and merge" option? (in the main branch, there will be just 1 `"Implement threshold cryptography"` commit, which will include all the changes from this PR). Otherwise, you would need to clean up the commit history and rebase it over the current `main` (as opposed to merging `main` into your branch).

abread:
> sorry that you had to do it several times already, but could you please merge the latest main branch into your branch?
Ah, it was already done (when I added the DSL stuff) :)

> Also, would it be OK if we use the "squash and merge" option? 
Fine by me!

xosmig:
> Ah, it was already done (when I added the DSL stuff) :)

Apparently, you will need to do it again.

image

abread:
Weird, it gives me "Already up to date".
I'll try to rebase it against `main`

abread:
Rebased to main, should be alright now!

xosmig:
> Weird, it gives me "Already up to date".

Yes, sorry, it was my fault, I had the wrong option selected in github and the button for merging was gray 😩 
I removed my comment after I figured that out, but, apparently, too late.
I'm just waiting for the final OK from @matejpavlovic before merging.

abread:
Ah dw, it was quick to do!

matejpavlovic:
Merging. Thanks for your contribution!!

PR #224, opened 2022-09-14
SMR System

matejpavlovic:
This PR introduces "Mir systems" (just a preliminary name, can be changed).
A system is a collection of Mir modules that are configured to work together, encapsulated in a single object.

The first such system is the SMR system. It is implemented in its own package and is expected to be used by the main function of a mir-based program.
It should replace the repetitive instantiation and configuration of all modules wherever a Mir node implementing SMR is created (in the test, in the chat app, in the debugger, ...).
So far only the chat app is using the SMR abstraction, while at other places the modules are still instantiated manually.
This should be changed in future PRs.


dnkolegov (approved):
LGTM from my perspective.
I ran `semgrep` to be sure that there are no low-hanging fruits and also read the changes that must be used in Eudico.

pkg/systems/smr/system.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/filecoin-project/mir/pkg/iss"
+	"github.com/filecoin-project/mir/pkg/logging"
+	"github.com/filecoin-project/mir/pkg/mempool/simplemempool"
+	"github.com/filecoin-project/mir/pkg/modules"
+	"github.com/filecoin-project/mir/pkg/net"
+	libp2pnet "github.com/filecoin-project/mir/pkg/net/libp2p"
+	t "github.com/filecoin-project/mir/pkg/types"
+)
+
+// System represents s Mir SMR system.
```
xosmig, line 24:
```suggestion
// System represents a Mir SMR system.
```

PR #230, opened 2022-09-20
Improve network transport

dnkolegov:


matejpavlovic (approved):
Good job! Most comments are just cosmetic.
Thanks! :-)

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -105,84 +113,122 @@ func (t *Transport) Stop() {
 	} else {
 		t.logger.Log(logging.LevelDebug, fmt.Sprintf("node %v libp2p host closed", t.ownID))
```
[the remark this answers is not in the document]
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -290,23 +344,84 @@ func (t *Transport) mirHandler(s network.Stream) {
...
+
+func (t *Transport) isConnectionToNodeInProgress(nodeID types.NodeID) bool {
+	t.nodesLock.RLock()
+	defer t.nodesLock.RUnlock()
+
+	info, found := t.nodes[nodeID]
+	return found && info.IsOpening
+}
+
+func (t *Transport) addConnectionInProgress(nodeID types.NodeID) {
```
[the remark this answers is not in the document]
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -105,84 +113,122 @@ func (t *Transport) Stop() {
...
+		t.logger.Log(logging.LevelError, fmt.Sprintf("failed to parse addr %v: %v", addr, err))
+		return
 	}
 
 	t.host.Peerstore().AddAddrs(info.ID, info.Addrs, PermanentAddrTTL)
 
 	s, err := t.openStream(ctx, info.ID)
 	if err != nil {
-		return nil, fmt.Errorf("failed to open new stream to node %v: %w", addr, err)
+		t.removeConnectionInProgress(node)
```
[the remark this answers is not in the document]
dnkolegov: Good catch! removed!

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -219,38 +265,46 @@ func (t *Transport) openStream(ctx context.Context, p peer.ID) (network.Stream,
...
 
-func (t *Transport) Send(dest types.NodeID, payload *messagepb.Message) error {
-	outBytes, err := proto.Marshal(payload)
-	if err != nil {
-		return fmt.Errorf("failed to marshal message: %w", err)
+func (t *Transport) Send(ctx context.Context, dest types.NodeID, payload *messagepb.Message) error {
+	found := t.nodeExists(dest)
+	if !found {
+		return fmt.Errorf("failed to get address for node %s", dest)
 	}
```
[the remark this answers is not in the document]
dnkolegov: The difference is that if the target node is not found it is not possible to add this node within libp2p. And because of that, it is not possible to open a stream. Only the caller can fix it somehow (e.g., via Connect function).

Does it make sense?

>  handled by t.connectToNode.
That's true. But `connectToNode` doesn't return an error. And because of that and async nature of that function, we will run
```
go t.connectToNode(ctx, dest, t.connWg)
		return fmt.Errorf("no open stream for node %s, reopening", dest)
```
Then we get an error "that node doesn't exist" in logs and get "no open stream for node" error in the caller function.

As a result, I think we should distinguish between cases when the node doesn't exist at all and the case when the node exists but the stream doesn't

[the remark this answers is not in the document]
matejpavlovic: Yes I agree that this potentially might be a useful thing to distinguish, but unless it happens atomically (checking both things while holding the lock), one can always construct a scenario where the first check is futile.
E.g., the check succeeds, releases the lock, and then someone removes the node by calling `CloseOldConnections`.

If we do want to distinguish it properly, I think the check still needs to happen atomically (for both the presence of the nodeInfo and the connection within). For example by extending the `t.getStream` function so we can make this distinction based on what it returns.

In fact, in any case, it still seems possible that the stream is closed *concurrently* by `CloseOldConnections` when `Send` is writing to it. It would be good to check with the doc that this doesn't panic and behaves in a predictable way.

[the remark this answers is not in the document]
dnkolegov: Agree. Fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -290,23 +344,84 @@ func (t *Transport) mirHandler(s network.Stream) {
...
+		return
 	}
-	t.outboundStreams[nodeID] = s
+	v.Stream = s
 }
 
-func (t *Transport) streamExists(nodeID types.NodeID) bool {
-	t.outboundStreamsMx.Lock()
-	defer t.outboundStreamsMx.Unlock()
+func (t *Transport) nodeExists(nodeID types.NodeID) bool {
```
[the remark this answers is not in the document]
dnkolegov: Fixed

matejpavlovic:
Nice, thanks for the fixes! Feel free to merge when you make the tests pass :-)

jsoares:
Continuation of Slack thread wrt to connection retries: we seem to have an answer wrt to libp2p behaviour (https://github.com/libp2p/go-libp2p/blob/7828f3e0797e0a7b7033fa5e8be9b94f57a4c173/p2p/net/swarm/swarm.go#L358, per Alfonso). Is the path clear on our end, or do we still need input from the libp2p team?

cc @dnkolegov @adlrocha @vukolic @sergefdrv

dnkolegov:
Yes, probably, we have an answer to the question.

>  Is the path clear on our end, or do we still need input from the libp2p team?

@jsoares I would be happy if someone from the libp2p team reviews this PR! Again, having in mind that Mir is going to the production.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,492 @@
...
+	}
+
+	if ctx.Err() != nil {
+		return err
+	}
+
+	w := bufio.NewWriter(s)
+	_, err = w.Write(buf.Bytes())
+	if err == nil {
+		w.Flush()
```
sergefdrv, line 326: IIUC, `Flush` may write the buffered data only partially and return `io.ErrShortWrite` ([source](https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/bufio/bufio.go;drc=99f1bf54eb502e123e60c605212a616146fbe86a;l=631)). Have you checked if libp2p stream writer may write less bytes than requested?

sergefdrv, line 326: This call will block if the underlying writer is busy. I'm afraid that any destination node could cause the whole module to block here.

dnkolegov: Yes, good catch. That is a debt from the initial implementation. Fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	if ctx.Err() != nil {
+		return err
+	}
+
+	w := bufio.NewWriter(s)
+	_, err = w.Write(buf.Bytes())
+	if err == nil {
+		if err = w.Flush(); err != nil {
```
sergefdrv, line 326: It may still write data partially to the libp2p stream. What happens on the other end in that case?

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,494 @@
...
+		var payload messagepb.Message
+
+		if err := proto.Unmarshal(msg.Payload, &payload); err != nil {
+			t.logger.Log(logging.LevelError, "failed to unmarshall mir request", "src", t.ownID, "dst", s.ID(), "err", err)
+			return
+		}
+
+		t.incomingMessages <- events.ListOf(
+			events.MessageReceived(types.ModuleID(payload.DestModule), types.NodeID(msg.Sender), &payload),
+		)
```
sergefdrv, line 370: What if the destination module does not exist?

pkg/net/libp2p/libp2p_transport.go
sergefdrv: I wonder if libp2p itself can write the data partially and return `io.ErrShortWrite` without the stream being broken

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Do you mean what if the destination module doesn't read from the channel?
Or what will Mir do if it gets a message to a module that doesn't exist?

sergefdrv: The latter

dnkolegov: @matejpavlovic Do you see any problems with that? I think that the network transport should know about that and assume that Mir properly handles that.

matejpavlovic: Yes, this is not the Net module's responsibility. The node must handle the situation. This is part of input sanitization.
While complex input is a bigger issue, I've just addressed this particular part in #253 . So I think this comment can be marked as resolved.

pkg/net/libp2p/libp2p_transport.go
matejpavlovic: Can a short write happen even without the connection breaking? If that's the case, we might need to forcefully break the connection and create a new one (or somehow else tell the other end that whatever comes next is a new message, not the continuation of the partially sent one)...

matejpavlovic: Yes, that is a good point too. We should make sure that a single slow connection doesn't block all the other ones. A proper solution to this might be actually non-trivial, probably involving separate per-connection output buffers, each operated by its own goroutine, and dropping outgoing messages when they become full... But maybe we can address this in a separate PR.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,500 @@
...
+				Addr:      addr,
+				IsOpening: false,
+				Stream:    nil,
+			}
+
+		}
+	}
+	t.nodesLock.Unlock()
+
+	t.connWg.Add(1)
```
adlrocha, line 187: Why are you using this `WaitGroup` here, instead of including all this waiting logic directly in `connect` (It feels harder to reason about if the waiting logic is scattered through different functions)? Can the high-level function `Connect` be called by several routines?

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,500 @@
...
+	go t.connect(ctx, nodes)
```
adlrocha, line 188: Do we really need this in a routine? We are already spawning new routines for each node when we call `connect`.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,500 @@
...
+	}
+
+	w := bufio.NewWriter(s)
+	_, err = w.Write(buf.Bytes())
+	if err == nil {
+		err = w.Flush()
+	}
+	// If we cannot write to the stream, or we don't have a connection to that peer then try to
+	// connect to the peer.
+	if err != nil || len(t.host.Network().ConnsToPeer(s.Conn().RemotePeer())) == 0 {
```
adlrocha, line 336: We should have a connection here by now, right? You already called `connectToNode` above so this is a sanity-check, but do we really want to try to connect again with the node if the stream failed? Maybe is just better to restart the connection (`Close() + Open()).

Also we should lock the stream while we write on it to prevent parallel writes. Instead of just locking in `getNodeStream` we should have a lock also while we write on it.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,500 @@
...
+	if !found {
+		return nil, &nodeInfoError{nodeID}
+	}
+
+	node, found := t.nodes[nodeID]
+	if !found ||
+		node.Stream == nil {
+		return nil, &nilStreamError{SrcNode: t.ownID, DestNode: nodeID}
+	}
+	return node.Stream, nil
```
adlrocha, line 412: As mentioned above, Streams should be locked while being written to avoid parallel accesses.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,500 @@
...
+						select {
+						case t.incomingMessages <- events.ListOf(receivedEvent):
+						case <-ctx.Done():
+						}
+					}()
+				} else {
+					// Send message to another node.
+					if err := t.Send(ctx, types.NodeID(destID), e.SendMessage.Msg); err != nil { // nolint
+						// TODO: Handle sending errors (and remove "nolint" comment above).
+						//       Also, this violates the non-blocking operation of ApplyEvents method. Fix it.
```
adlrocha, line 490: I don't get this. Isn't this a "fire-and-forget"? We don't wait for a response, do we?

pkg/net/libp2p/libp2p_transport_test.go
```diff
@@ -0,0 +1,309 @@
...
+	t.Log("membership")
+	t.Log(m.membership)
+
+	initialNodes := m.Membership(nodeA, nodeB, nodeC)
+
+	a.Connect(ctx, initialNodes)
+	b.Connect(ctx, initialNodes)
+	c.Connect(ctx, initialNodes)
+
+	m.testEventuallyConnected(nodeA, nodeB)
```
adlrocha, line 281: Should we also test if the retries work and have some parallel tests (with calls from different routines)? I haven´t looked in depth into the code, so maybe they are already there.

adlrocha (changes requested):
Overall, LGTM. Just a request for change and a suggestion:
- We need to lock writes to the streams so several routines don´t try to write at the same time to the same stream.
- It is a bit hard to reason about the correctness of the concurrency in the connection manager. I wouldn't recommend having an active `WaitGroup` scattered through different functions if we can avoid it. It makes the code harder to read and it may lead to deadlocks (but maybe is just personal taste, so take this with a grain of salt).

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Yes, good catch! I combined two functions and forgot to change locking

pkg/net/libp2p/libp2p_transport.go
dnkolegov: We need if we want to not wait when all connecting goroutines stop, otherwise `wg.Wait` blocks us.

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Agree, I moved this inside of `connect` function

pkg/net/libp2p/libp2p_transport.go
dnkolegov: > You already called connectToNode above

That was called in case we don't have a stream to write in. 
But what if we have already had a stream, we tried to write in it, and get an error?
But I agree we should close the connection!

pkg/net/libp2p/libp2p_transport_test.go
matejpavlovic: (This applies to some of the previous comments too.)

In the current implementation, there is no way a connection could be written to concurrently by multiple threads. The whole module is handled by a single thread, sending messages one by one. This is also why getting stuck on writing to one broken (or slow) connection is such a big problem. However, at least we don't have to be concerned about locking the streams when writing into them (unless we need to synchronize the writes with potential reconnections).

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed. It was legacy code

pkg/net/libp2p/libp2p_transport_test.go
dnkolegov: We have a basic check that retry works - https://github.com/filecoin-project/mir/blob/be26a6d97e5d97d0c2a87ba4056f5d16cfc0a58f/pkg/net/libp2p/libp2p_transport_test.go#L235

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Created an issue to address this - https://github.com/filecoin-project/mir/issues/255

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Created an issue to address this - https://github.com/filecoin-project/mir/issues/255

dnkolegov: Created an issue to address this - https://github.com/filecoin-project/mir/issues/255

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,463 @@
...
+func (t *Transport) streamExists(nodeID types.NodeID) bool {
+	t.nodesLock.RLock()
+	defer t.nodesLock.RUnlock()
+
+	v, found := t.nodes[nodeID]
+	return found && v.Stream != nil
+}
+
+func (t *Transport) getNodeStreamAndInfo(nodeID types.NodeID) (network.Stream, *peer.AddrInfo, error) {
+	t.nodesLock.Lock()
```
adlrocha, line 363: This is not enough, you need to keep the lock while you write the stream. I would:
- Make this function `getNodeStreamAndInfoWithoutLock`.
- Use this function when getting the stream to write by locking before getting and releasing after flush.
- Create a `getNodeStreamAndInfoWithLock` and keep the `RLock` you had before.

Reference: https://github.com/filecoin-project/mir/pull/230/commits/1667e25c53c50b79e0ab4c9d2a9bb9ba5b4284a0

pkg/net/libp2p/libp2p_transport.go
adlrocha: I don't know IIU the issue correctly, but I honestly don't think this is possible. Libp2p expects the buffer to be explicitly flushed so it can notify the length of the stream being sent to other side (all stream communications are prefixed by a `VarInt` notifying the receiver the amount of bytes to read from the buffer, and buffers are reused on top of the existing connections).

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,484 @@
...
+	stream, err := t.getNodeStreamAndInfoWithoutLock(dest)
+	if err != nil {
+		return err
+	}
+
+	w := bufio.NewWriter(stream)
+	if _, err = w.Write(payload); err != nil {
+		return err
+	}
+	return w.Flush()
```
sergefdrv, line 294: What is the advantage of using `bufio.Writer` here?

pkg/net/libp2p/libp2p_transport.go
sergefdrv: I think one can make an argument that it's the module's fault emitting events which target a wrong module.

pkg/net/libp2p/libp2p_transport.go
dnkolegov: That is a common [approach](https://github.com/libp2p/go-libp2p/tree/master/examples) for libp2p applications.

dnkolegov: What would you recommend to use instead of that?

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,484 @@
...
+		}
+	}
+	t.nodesLock.Unlock()
+
+	t.connect(ctx, nodes)
+}
+
+func (t *Transport) connect(ctx context.Context, nodes map[types.NodeID]types.NodeAddress) {
+	t.connWg.Add(1)
+	defer t.connWg.Done()
```
sergefdrv, line 163: This pair of lines doesn't seem to make sense: the subsequent go statement will return right away causing the defer statement to execute immediately.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,484 @@
...
+func (t *Transport) connectToNode(ctx context.Context, node types.NodeID, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	if t.isConnectionToNodeInProgress(node) {
+		t.logger.Log(logging.LevelDebug, "connecting to node is in progress", "src", t.ownID, "dst", node)
+		return
+	}
+
+	t.setConnectionInProgress(node)
+	defer t.clearConnectionInProgress(node)
```
sergefdrv, line 213: Are those calls supposed to make some synchronization scheme up or they are just there for reducing the likelihood of duplicate attempt to establish the same connection? I think there is a possible race condition since `t.nodesLock` gets unlocked in-between.

pkg/net/libp2p/libp2p_transport.go
sergefdrv: Why not writing to the libp2p stream directly?

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,484 @@
...
+	}() // nolint
+
+	for {
+		msg, sender, err := t.decode(s)
+		if err != nil {
+			t.logger.Log(logging.LevelError, "failed to read mir request", "src", t.ownID, "sender", sender, "err", err)
+			return
+		}
+		t.incomingMessages <- events.ListOf(
+			events.MessageReceived(types.ModuleID(msg.DestModule), sender, msg),
```
sergefdrv, line 363: I don't think we can trust `sender` here, since we receive it from the other, potentially faulty, node and doesn't seem to verify in any way.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,484 @@
...
+}
+
+var _ mirnet.Transport = &Transport{}
+
+var ErrUnknownNode = errors.New("unknown node")
+var ErrNilStream = errors.New("stream has not been opened")
+
+type nodeInfo struct {
+	ID        types.NodeID
+	Addr      types.NodeAddress
```
sergefdrv, line 51: `ID` and `Addr` don't seem to be used.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,484 @@
...
+				continue
+			}
+
+			if t.streamExists(nodeID) {
+				t.logger.Log(logging.LevelInfo, "stream to node already exists", "src", t.ownID, "dst", nodeID)
+				wg.Done()
+				continue
+			}
+
+			go t.connectToNode(ctx, nodeID, wg)
```
sergefdrv, line 185: `connectToNode` doesn't seem to re-check if the stream is already exist with `nodesLock` held. I think there is a possible race condition.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,484 @@
...
+
+	go func() {
+		defer t.connWg.Done()
+
+		t.nodesLock.Lock()
+		defer t.nodesLock.Unlock()
+
+		for nodeID, dest := range t.nodes {
+			// Close an old connection to a node if we don't need to connect to it further.
+			if _, foundInNewNodes := newNodes[nodeID]; !foundInNewNodes {
```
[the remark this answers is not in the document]
dnkolegov: I don't see a reason to do it in the current network addressing in Mir since all nodes are identified by NodeIDs.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,484 @@
...
+	if err := t.host.Close(); err != nil {
+		t.logger.Log(logging.LevelError, "Could not close libp2p", "ownID", t.ownID, "err", err)
+	} else {
+		t.logger.Log(logging.LevelDebug, "libp2p host closed", "ownID", t.ownID)
+	}
+
+	t.connWg.Wait()
+}
+
+func (t *Transport) CloseOldConnections(newNodes map[types.NodeID]types.NodeAddress) {
```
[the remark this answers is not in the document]
dnkolegov: @matejpavlovic what do you think? do we need this?

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed. `defer t.connWg.Done()` must be in the goroutine

pkg/net/libp2p/libp2p_transport.go
dnkolegov: I can do the following:
```
                msg, sender, err := t.decode(s)
		if err != nil {
			t.logger.Log(logging.LevelError, "failed to read mir request", "src", t.ownID, "sender", sender, "err", err)
			return
		}
		t.nodesLock.RLock()
		v, ok := t.nodes[sender]
		t.nodesLock.RUnlock()
		if !ok {
			return
		}
		msgSenderID := v.AddrInfo.ID.String()
		senderPeerID := s.Conn().RemotePeer().String()
		if msgSenderID != senderPeerID {
			t.logger.Log(logging.LevelWarn, 
				"failed to validate sender",
				"src", t.ownID, "sender", sender, "msgSender", msgSenderID, "peerID", senderPeerID)
			return
		}
		t.incomingMessages <- events.ListOf(
			events.MessageReceived(types.ModuleID(msg.DestModule), sender, msg),
		)
```

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Do you want to check that again?

pkg/net/libp2p/libp2p_transport.go
dnkolegov: fixed

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed

sergefdrv: Now we have some additional overhead for every received message. We may want to improve this in future.

pkg/net/libp2p/libp2p_transport.go
sergefdrv: Not necessarily, we just need to make sure no race condition happens here.

pkg/net/libp2p/libp2p_transport.go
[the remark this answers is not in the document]
dnkolegov: We don't support this case now. But if we did then we would need to answer the following question: what if we have a connection to nodeID and receive the same nodeID with a different address? what should we do? Do we need to establish a new connection and remove the old one, to support both? Until we don't have a clear requirement and use case for that let's use the simplest approach.

adlrocha:
Quick question before we land this PR. While integrating Mir I realized that we are using a single node address to contact peers, but peers (and specifically libp2p hosts) may expose more than one endpoint. Do we want to support this? (I think we should but I don't know how hard it is to include, if it is hard lets track on an issue and defer to the future). 

@matejpavlovic @dnkolegov thoughts? 🙏

dnkolegov:
> Quick question before we land this PR. While integrating Mir I realized that we are using a single node address to contact peers, but peers (and specifically libp2p hosts) may expose more than one endpoint. Do we want to support this? (I think we should but I don't know how hard it is to include, if it is hard lets track on an issue and defer to the future).
> 
> @matejpavlovic @dnkolegov thoughts? 🙏

I would prefer to defer to the future. The current implementation assumes basic Mir's membership model - nodeID:nodeAddress

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed

dnkolegov:
@sergefdrv @matejpavlovic @adlrocha 

I simplified the implementation.
Now Connect, CloseOldConnections, and reconnect can not be run simultaneously. But Send can work in parallel with all these functions. Let me know what you think.

Could you also resolve the conversations if you believe I have implemented your suggestions or if it is not actual anymore?

pkg/net/libp2p/libp2p_transport.go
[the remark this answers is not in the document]
sergefdrv: Right, we don't yet support updating the validator network address in the subnet actor. I think we should reverse the direction of communication between nodes such that the node receives messages only from those peers which it initiated connections to. Similarly, incoming connections would be served by primarily supplying messages there. Nodes could reject incoming connections from nodes with unexpected identities.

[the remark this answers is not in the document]
sergefdrv: > what if we have a connection to nodeID and receive the same nodeID with a different address? what should we do? Do we need to establish a new connection and remove the old one, to support both?

I think we should interpret this as a simple network endpoint update and replace the old connection with a new one.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+
+var _ mirnet.Transport = &Transport{}
+
+var ErrUnknownNode = errors.New("unknown node")
+var ErrNilStream = errors.New("stream has not been opened")
+
+type connInfo struct {
+	AddrInfo  *peer.AddrInfo
+	Stream    network.Stream
+	IsOpening bool
```
sergefdrv, line 51:
```suggestion
```

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+
+		t.connWg.Add(1)
+		go func() {
+			defer t.connWg.Done()
+
+			t.connsLock.Lock()
+			defer t.connsLock.Unlock()
+
+			t.connWg.Add(1)
+			t.connectToNodeWithoutLock(ctx, node, t.connWg)
```
sergefdrv, line 268: Now that you don't start a new goroutine here, this usage of `t.connWg` looks superfluous.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+	t.host.RemoveStreamHandler(ProtocolID)
+
+	t.connsLock.Lock()
+	for nodeID, addr := range t.conns {
+		if err := t.host.Network().ClosePeer(addr.AddrInfo.ID); err != nil {
+			t.logger.Log(logging.LevelError, "could not close old connection to node", "src", t.ownID, "dst", nodeID, "err", err)
+			continue
+		}
+	}
+	t.connsLock.Unlock()
```
sergefdrv, line 98: Does this prevent reconnection right afterwards?

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+}
+
+func (t *Transport) reconnect(ctx context.Context, node types.NodeID) {
+	info, found := t.getNodeAddrInfo(node)
+	if !found {
+		t.logger.Log(logging.LevelError, "failed to get node address", "src", t.ownID, "dst", node)
+	} else {
+		if err := t.host.Network().ClosePeer(info.ID); err != nil {
+			t.logger.Log(logging.LevelError, "could not close the connection to node", "src", t.ownID, "dst", node, "err", err)
+		}
```
sergefdrv, line 258: Perhaps we should also protect this with `t.connsLock`

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+
+	// The implementation is based on the openStream() function from the RemoteTracer:
+	// https://github.com/libp2p/go-libp2p-pubsub/blob/cbb7bfc1f182e0b765d2856f6a0ea73e34d93602/tracer.go#L280
+	var s network.Stream
+	var err error
+	for i := 0; i < maxRetries; i++ {
+		sctx, cancel := context.WithTimeout(ctx, maxConnectingTimeout)
+
+		s, err = t.host.NewStream(sctx, dest, ProtocolID)
+		cancel()
```
sergefdrv, line 316: Why do you call `cancel` here?

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+				"src", t.ownID, "sender", sender, "senderNodeID", senderNodeID, "senderPeerID", senderPeerID)
+			return
+		}
+		t.incomingMessages <- events.ListOf(
+			events.MessageReceived(types.ModuleID(msg.DestModule), sender, msg),
+		)
+	}
+}
+
+func (t *Transport) getNodeStreamAndInfoWithoutLock(nodeID types.NodeID) (network.Stream, error) {
```
sergefdrv, line 431:
```suggestion
func (t *Transport) getNodeStreamWithoutLock(nodeID types.NodeID) (network.Stream, error) {
```

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+	t.logger.Log(logging.LevelDebug, "starting libp2p transport", "ownID", t.ownID, "listen", t.host.Addrs())
+	t.host.SetStreamHandler(ProtocolID, t.mirHandler)
+	return nil
+}
+
+func (t *Transport) Stop() {
+	t.logger.Log(logging.LevelDebug, "stopping libp2p transport", "ownID", t.ownID)
+	defer t.logger.Log(logging.LevelDebug, "stopping libp2p transport finished.", "ownID", t.ownID)
+
+	t.host.RemoveStreamHandler(ProtocolID)
```
sergefdrv, line 89: Does this also close incoming connections?

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Because that is context.WithTimeout's cancel. 
```
Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this
```

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
dnkolegov: It doesn't close connections directly
https://github.com/multiformats/go-multistream/blob/225d479012e89db2156d325a60b9374713f21769/multistream.go#L148

pkg/net/libp2p/libp2p_transport.go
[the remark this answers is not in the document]
dnkolegov: Then I would suggest addressing it later.
https://github.com/filecoin-project/mir/issues/259

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Agree, but we use the same function and to do that we need a wg.
And we can't just call `go t.connectToNodeWithoutLock(ctx, node, t.connWg)` due to Lock/Unlock

pkg/net/libp2p/libp2p_transport.go
dnkolegov: No, it doesn't. If you call `Stop` and then call `Connect` then new outbound connections can be established but new inbound connections to Mir protocol can't.
We can't close the `host` since we don't own it.

pkg/net/libp2p/libp2p_transport.go
dnkolegov: t.connsLock is used to protect t.conns. `t.host.Network().ClosePeer` doesn't change or read `t.conns`elements

pkg/net/libp2p/libp2p_transport.go
sergefdrv: Then I think we'd better close all the streams in `mirHandler`.

pkg/net/libp2p/libp2p_transport.go
dnkolegov: No. `t.conns` is just information about address and streams

pkg/net/libp2p/libp2p_transport.go
dnkolegov: `mirHandler` is the same as a handler in`http`. I don't see how to implement that.

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Added removing address from t.conns when calling `Stop`

sergefdrv: Well, we get `t.conns[node].Stream` from `t.host`. I was wondering if calling `t.host.Network().ClosePeer` invalidates the corresponding stream as a side-effect.

pkg/net/libp2p/libp2p_transport.go
sergefdrv: It receives `network.Stream` which has `Close` method. I think we should arrange all loops running in `mirHandler` to call `s.Close` once we call `Stop`.

sergefdrv: The difference from an HTTP handler is that `mirHandler` has a long-running loop inside. If we stop the transport module, it should also drop the incoming connections, otherwise the corresponding resources cannot be garbage-collected.

pkg/net/libp2p/libp2p_transport.go
sergefdrv: I mean `Stop` and `reconnect` methods running concurrently.

pkg/net/libp2p/libp2p_transport.go
sergefdrv: Ok, I see that `sendPayload` and `reconnect` cannot run concurrently for the same node.

pkg/net/libp2p/libp2p_transport.go
dnkolegov: That shows that it doesn't close connections.

dnkolegov: Probably, I did not understand you right.

sergefdrv: I suspect that neither `RemoveStreamHandler` nor `Network().ClosePeer` would close an already established incoming stream connection.

sergefdrv: Stream and/or connection :)

sergefdrv: > ClosePeer closes the connection to a given peer

Here I'm not sure it also closes incoming connections.

adlrocha: FWIW (and I just skimmed through the discussion), I usually close all streams before closing the host as a sanity check with something like:
```go 
// Unregister handler
...
// Close streams
for _, s := range streams{
   s.Close()
}
// Close host
h.Close()
```

dnkolegov: If I understood Sergay correctly he suggests closing all streams in the handler. 
Currently, the handler closes the stream on defer.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+		wg.Add(len(nodes))
+
+		for nodeID := range nodes {
+			// Do not establish a real connection with own node.
+			if nodeID == t.ownID {
+				wg.Done()
+				continue
+			}
+
+			go t.connectToNodeWithoutLock(ctx, nodeID, wg)
```
adlrocha, line 245: I think we are missing a `wg.Done()` below this line.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,452 @@
...
+		t.connsLock.Lock()
+		defer t.connsLock.Unlock()
+
+		t.connWg.Add(1)
+		t.connectToNodeWithoutLock(ctx, node, t.connWg)
+	}()
+}
+
+func (t *Transport) connectToNodeWithoutLock(ctx context.Context, nodeID types.NodeID, wg *sync.WaitGroup) {
+	defer wg.Done()
```
adlrocha, line 274:
```suggestion
	defer wg.Done()
```

This is not needed anymore right? You handle the waiting groups in the functions where they are called, mainly `reconnect` and `connect`

adlrocha (approved):
Ready to merge!

pkg/net/libp2p/libp2p_transport.go
sergefdrv: Yes, on `Stop`, I suggest closing all the streams received in `mirHandler` and waiting until all handlers exit.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,459 @@
...
+				continue
+			}
+
+			go t.connectToNodeWithoutLock(ctx, nodeID, wg)
+		}
+		wg.Wait()
+	}()
+}
+
+func (t *Transport) reconnect(ctx context.Context, node types.NodeID) {
```
matejpavlovic, line 249: Related to previous comment: Is there a fundamental difference between `connect` and `reconnect`, except the former does it for multiple nodes in a batch?

pkg/util/libp2p/libp2p.go
```diff
@@ -31,6 +32,24 @@ func NewDummyHost(numericID int, sourceAddr t.NodeAddress) host.Host {
...
 
+func NewFreeHostAddr() (m multiaddr.Multiaddr) {
+	var a *net.TCPAddr
+	var err error
+	a, err = net.ResolveTCPAddr("tcp", "localhost:0")
+	if err == nil {
+		var l *net.TCPListener
+		if l, err = net.ListenTCP("tcp", a); err == nil {
+			defer l.Close() // nolint
+			addr := l.Addr().(*net.TCPAddr).Port
```
matejpavlovic, line 43: Cosmetic: Calling a port `addr` is a bit deceiving.

matejpavlovic (commented):
I added some more comments, but some of them I suggest to address in the next PR. Will att them to #255 .

pkg/net/libp2p/libp2p_transport.go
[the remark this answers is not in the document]
dnkolegov: Will be addressed [here](https://github.com/filecoin-project/mir/issues/255). The most important question is to define the requirement!

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,459 @@
...
+			}
+			t.logger.Log(logging.LevelError, "failed to read mir message", "src", t.ownID, "err", err)
+			return
+		}
+		t.connsLock.RLock()
+		senderNode, ok := t.conns[sender]
+		t.connsLock.RUnlock()
+		if !ok {
+			return
+		}
```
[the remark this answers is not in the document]
dnkolegov: I don't think so. It even doesn't return if you close the host.

pkg/util/libp2p/libp2p.go
dnkolegov: fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,459 @@
...
+			continue
+		}
+		_, found := t.conns[nodeID]
+		if !found {
+			t.conns[nodeID] = &connInfo{AddrInfo: parsedAddrInfo[nodeID], Stream: nil}
+		}
+	}
+	t.connsLock.Unlock()
+
+	t.connect(ctx, nodes)
```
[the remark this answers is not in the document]
dnkolegov: Fixed.

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,459 @@
...
+	defer func() {
+		t.logger.Log(logging.LevelDebug, "mir handler is closing stream", "src", t.ownID, "dst", peerID)
+		err := s.Close()
+		if err != nil {
+			t.logger.Log(logging.LevelError, "closing stream", "src", t.ownID, "dst", peerID, "err", err)
+		}
+	}() // nolint
+
+	for {
+		msg, sender, err := t.readAndDecode(s)
```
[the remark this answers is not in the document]
dnkolegov: What "preemptable call" is? The call using gosched?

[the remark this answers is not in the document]
dnkolegov: Added stopChan. Did you mean something like that? if you did not let's discuss

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,459 @@
...
+	buf := new(bytes.Buffer)
+	if err = tm.MarshalCBOR(buf); err != nil {
+		return nil, fmt.Errorf("failed to CBOR marshal message: %w", err)
+	}
+	return buf.Bytes(), nil
+}
+
+func (t *Transport) readAndDecode(s io.Reader) (*messagepb.Message, types.NodeID, error) {
+	var tm TransportMessage
+	err := tm.UnmarshalCBOR(s)
```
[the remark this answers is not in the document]
dnkolegov: Addressed via `stopChan`

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,459 @@
...
+		senderNodeID := senderNode.AddrInfo.ID.String()
+		senderPeerID := s.Conn().RemotePeer().String()
+		if senderNodeID != senderPeerID {
+			t.logger.Log(logging.LevelWarn,
+				"failed to validate sender",
+				"src", t.ownID, "sender", sender, "senderNodeID", senderNodeID, "senderPeerID", senderPeerID)
+			return
+		}
+
+		t.incomingMessages <- events.ListOf(
```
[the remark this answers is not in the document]
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,459 @@
...
+		wg.Add(len(nodes))
+
+		for nodeID := range nodes {
+			// Do not establish a real connection with own node.
+			if nodeID == t.ownID {
+				wg.Done()
+				continue
+			}
+
+			go t.connectToNodeWithoutLock(ctx, nodeID, wg)
```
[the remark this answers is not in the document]
dnkolegov: If I remember correctly the main requirement is that connection and sending must be nonblocking. Correct?
If correct then we need a goroutine over connecting goroutines, otherwise, this call will be blocking.
If you meant by "nonblocking" something else then define it and I will reimplement it.

sergefdrv (approved):

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,454 @@
...
+			case one := <-t.streamChan:
+				t.connsLock.Lock()
+				conn, found := t.conns[one.NodeID]
+				if !found {
+					t.connsLock.Unlock()
+					continue
+				}
+				conn.Stream = one.Stream
+				t.connsLock.Unlock()
+				t.logger.Log(logging.LevelDebug, "updated stream", "src", t.ownID, "nodeID", one.NodeID)
```
matejpavlovic, line 247: If we found the stream, let's try to close the old one (after releasing the lock).

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,454 @@
...
+	if conn.Stream != nil && t.host.Network().Connectedness(conn.AddrInfo.ID) == network.NotConnected {
+		if err := conn.Stream.Close(); err != nil {
+			t.logger.Log(logging.LevelError, "could not close stream to node", "src", t.ownID, "dst", nodeID, "err", err)
+		}
+	}
+
+	info := conn.AddrInfo
+
+	t.host.Peerstore().AddAddrs(info.ID, info.Addrs, PermanentAddrTTL)
+
```
matejpavlovic, line 288: Release RLock here.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,454 @@
...
+	for i := range nodesID {
+		// Do not establish a real connection with own node.
+		if nodesID[i] == t.ownID {
+			continue
+		}
+		go t.connectToNodeWithoutLock(ctx, nodesID[i])
+	}
+}
+
+func (t *Transport) connectToNodeWithoutLock(ctx context.Context, nodeID types.NodeID) {
```
matejpavlovic, line 263: Pls also rename this (e.g. to `connectToNode`) - now it's using a lock again :-)

dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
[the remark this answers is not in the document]
matejpavlovic: Yes, by "preemptable" I meant that we must be able to interrupt the waiting for the data. The `stopChan` is a way of achieving that (since when `stopChan` is closed, the stream is closed, which makes the unmarshalling return, which in turn makes `readAndDecode` return).

pkg/net/libp2p/libp2p_transport.go
dnkolegov: Fixed

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	out, err := t.encode(msg)
+	if err != nil {
+		return err
+	}
+
+	err = t.sendPayload(dest, out)
+	if err != nil {
+		t.connect(ctx, []types.NodeID{dest})
```
matejpavlovic, line 176: I'm a bit worried that, if the connection breaks and the new one takes some time to establish, there might be literally thousands of messages that fail to be sent in the meantime, each of them triggering a new connection attempt. All of these attempts (but one) will be closed almost immediately. Could that happen?

matejpavlovic: In fact, nevermind, let's do this in a separate PR.

pkg/net/libp2p/libp2p_transport.go
```diff
@@ -0,0 +1,450 @@
...
+
+	t.connsLock.Lock()
+	for nodeID, c := range t.conns {
+		if err := c.Stream.Close(); err != nil {
+			t.logger.Log(logging.LevelError, "could not close stream", "src", t.ownID, "dst", nodeID, "err", err)
+		}
+		delete(t.conns, nodeID)
+	}
+	t.connsLock.Unlock()
+}
```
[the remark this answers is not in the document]
dnkolegov: Why do we need wg if we have the stopChan and it cannot exit until it gets a signal on this channel?

PR #232, opened 2022-09-24
Introduce fuzzing

dnkolegov:


matejpavlovic (approved):
Nice! I see this as a fuzzing "Hello World" to be expanded on later, for other parts of the code. Thanks!

PR #251, opened 2022-09-30
Event backpressure

### LINKED ISSUE #100: Event loop flow control
Implement a mechanism that monitors the state of event buffers in the `Node`'s main event loop and, if the buffers exceed a configured threshold, stop reading events from `ActiveModule`s.
For more details see discussion #51 .

### LINKED ISSUE #238: Finite buffers for incoming events from ActiveModules
Candidates: @matejpavlovic 
Estimate: 1wk

matejpavlovic:
This PR implements backpressure for event processing within a node.

Active modules now cannot make the node's event buffers explode in size, as the node stops reading input from active modules when its internal event buffer grows above a configurable limit.

The limit on the internal event buffer is a "soft limit" in the sense that as long as there are fewer events than the limit in the buffer, an arbitrary event list can be added, which might result in the buffer's actual size "overshooting" the limit, if the event list being added is long. This should not be a big problem in practice though, as the list already occupies space in memory and the events are simply "moved" to the buffer, without creating another copy.

Resolves #100 .
Resolves #238 .

### DIFF
--- a/eventbuffer.go
+++ b/eventbuffer.go
@@ -17,9 +17,9 @@ type eventBuffer struct {
 }
 
 // newEventBuffer allocates and returns a pointer to a new eventBuffer object.
-func newEventBuffer(modules modules.Modules) *eventBuffer {
+func newEventBuffer(modules modules.Modules) eventBuffer {
 
-	wi := &eventBuffer{
+	wi := eventBuffer{
 		buffers:     make(map[t.ModuleID]*events.EventList),
 		totalEvents: 0,
 	}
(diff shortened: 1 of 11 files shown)


eventbuffer.go
```diff
@@ -10,15 +10,21 @@ import (
...
 // It contains a separate list for each module.
-type eventBuffer map[t.ModuleID]*events.EventList
+type eventBuffer struct {
+	buffers     map[t.ModuleID]*events.EventList
+	totalEvents int
+}
 
 // newEventBuffer allocates and returns a pointer to a new eventBuffer object.
-func newEventBuffer(modules modules.Modules) eventBuffer {
+func newEventBuffer(modules modules.Modules) *eventBuffer {
```
sergefdrv, line 19: The `eventBuffer` structure doesn't seem to have grown a lot, why returning it by reference?

sergefdrv (approved):
Really nicely structured PR! Looks good, modulo a small comment :+1:

eventbuffer.go
matejpavlovic: It turned out that the `AddEvents` method must have a pointer receiver, otherwise a copy of the event counter would be incremented instead of the counter itself (it cost me some debugging time to figure that out... 🙄 ), so I just switched everything to pointers.

But indeed that's not necessary, so switching back to value, only keeping that method receiver as a pointer.

PR #253, opened 2022-10-03
Drop messages with non-existent destination module

matejpavlovic:
Related to the [PR comment about input sanitization](https://github.com/filecoin-project/mir/pull/230#discussion_r985320451).

Trivial change, merging without review.

### DIFF
--- a/eventbuffer.go
+++ b/eventbuffer.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/filecoin-project/mir/pkg/events"
 	"github.com/filecoin-project/mir/pkg/modules"
+	"github.com/filecoin-project/mir/pkg/pb/eventpb"
 	t "github.com/filecoin-project/mir/pkg/types"
 )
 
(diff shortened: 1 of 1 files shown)


eventbuffer.go
```diff
@@ -37,7 +38,13 @@ func (wi eventBuffer) AddEvents(events *events.EventList) error {
...
 			buffer.PushBack(event)
 		} else {
-			return errors.Errorf("no buffer for module %v (adding event of type %T)", event.DestModule, event.Type)
+			if _, ok := event.Type.(*eventpb.Event_MessageReceived); !ok {
+				// If the event is a MessageReceived event, we don't need to return an error, because
+				// the message may have been sent by a faulty node.
+				// MessageReceived events with non-existent module destinations are thus dropped,
+				// without it being considered an error at the local node.
+				return errors.Errorf("no buffer for module %v (adding event of type %T)", event.DestModule, event.Type)
+			}
```
sergefdrv, line 47: I think we'll need to take further measures to disconnect from faulty nodes and prevent DoS attacks.

matejpavlovic: Yes, absolutely, this is not meant to be an exhaustive solution.

sergefdrv: Do you have a rough idea of how to approach the issue?

matejpavlovic: Ultimately, we need to follow all code paths where received messages are processed and perform the necessary checks. Apart from, maybe introducing some limit on message size (probably directly in the network module), this boils down to sanitizing each module's input. At node level, there doesn't seem to be more that can be done, as the only thing a node does is to forward the message to the appropriate module (and the only field it looks at is `destModule` checked here).

matejpavlovic: Yes, this follows the principle of preferring pulling over pushing, which, in general I fully agree with. We definitely should keep this in mind for the next overhaul of the transport layer.

PR #257, opened 2022-10-05
Reverting libp2p upgrade

adlrocha:
- Lotus is still using `go-libp2p v0.20.3` that doesn't reflect the deprecation of `go-libp2p-core`. In order to use Mir in Lotus we need to revert to a previous of libp2p.


dnkolegov (approved):
LGTM!

PR #260, opened 2022-10-06
Publish checkpoints to application

matejpavlovic:
Publishing checkpoints to the application.

The last commit is a dirty one and only contains 2 tests, the difference between them only being that one uses simulated time and the other one doesn't. It seems that the one with simulated time doesn't properly handle `TimerRepeat` events.
I added some debug output to make the problem visible.

@sergefdrv Could you pls have a look at it? It looks like the issue might be in the simulated event repetition... Thanks!

Apart from that, @adlrocha this is probably all you need regarding interfacing with the integration.


matejpavlovic:
> Have you tried git-bisecting the issue?
There was no need for that, I knew the commit that introduced the problem, just didn't know how :-)
But that's all fixed now, will merge after rebasing on main.
Thanks for the help!

PR #261, opened 2022-10-06
Support module replacement in SMR system

matejpavlovic:


adlrocha (approved):
LGTM. I was thinking if it was worth making it a `WithModules` and support a list of modules as an argument, but I feel chaining is fine by now (I don't expect someone to replace more than one or two nodes at this point)

PR #277, opened 2022-10-18
Improve module error propagation

abread:
I accidentally sent `SendMessage` events to the threshcrypto module and got a cryptic error message.

This PR fixes error propagation in `ApplyEventsConcurrently`, and aborts simulations when modules produce errors.

Aborting sounds like the appropriate behavior because right now the error I described (sending events to the wrong module) will just quietly exit the simulated process and hang the test until it timeouts.


matejpavlovic (approved):
👍

PR #278, opened 2022-10-18
threshcrypto: Don't panic when no errors occur

abread:
When operations succeed, `err` is `nil`, and the module would panic when trying to call `err.Error()`


matejpavlovic (approved):
👍

PR #279, opened 2022-10-18
Start node from arbitrary checkpoint

### LINKED ISSUE #248: Restarting from arbitrary checkpoints

### LINKED ISSUE #276: Convenient checkpoint verification interface
Encapsulate Trantor stable checkpoints in a convenient interface that exposes functions for easily verifying the checkpoint certificate.

matejpavlovic:
Resolves #248 .
Resolves #276 .

### DIFF
--- a/pkg/checkpoint/checkpoint.go
+++ b/pkg/checkpoint/checkpoint.go
@@ -1,31 +1,95 @@
 package checkpoint
 
 import (
+	"fmt"
+
+	"github.com/filecoin-project/mir/pkg/crypto"
 	"github.com/filecoin-project/mir/pkg/pb/checkpointpb"
 	"github.com/filecoin-project/mir/pkg/pb/commonpb"
+	"github.com/filecoin-project/mir/pkg/serializing"
 	t "github.com/filecoin-project/mir/pkg/types"
 )
 
+// StableCheckpoint represents a stable checkpoint.
 type StableCheckpoint checkpointpb.StableCheckpoint
 
+// SeqNr returns the sequence number of the stable checkpoint.
+// It is defined as the number of sequence numbers comprised in the checkpoint, or, in other words,
+// the first (i.e., lowest) sequence number not included in the checkpoint.
 func (sc *StableCheckpoint) SeqNr() t.SeqNr {
 	return t.SeqNr(sc.Sn)
 }
 
+// Memberships returns the memberships configured for the epoch of this checkpoint
+// and potentially several subsequent ones.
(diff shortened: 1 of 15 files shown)


pkg/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/filecoin-project/mir/pkg/crypto"
+	"github.com/filecoin-project/mir/pkg/pb/checkpointpb"
+	"github.com/filecoin-project/mir/pkg/pb/commonpb"
+	"github.com/filecoin-project/mir/pkg/serializing"
+	t "github.com/filecoin-project/mir/pkg/types"
+)
+
+// StableCheckpoint represents a stable checkpoint.
+type StableCheckpoint checkpointpb.StableCheckpoint
```
adlrocha, line 14: Is there a `NewStableCheckpoint` function that goes from the arguments of a checkpoint to an instantiation of `StableCheckpoint`? This could be really handy to perform type translations between different data structures.

adlrocha (approved):
Overall, LGTM. If we get to a point where we are not time-pressured I would love to revisit how we handle types. It would be great if we can use abstractions over the protobufs decoupling a bit types from their underlying serialized representations, but that's a horse of another color :)

pkg/checkpoint/checkpoint.go
matejpavlovic: I see the point and I agree, but since the checkpoint is a rather complex data structure that is composed of different parts created at completely different times and places (that are protobufs themselves), unless I changed all that, the arguments would still need to be protobufs. I can add the `NewStableCheckpoint` function just to have it there already, but dealing with protobufs for now is almost inevitable.

matejpavlovic:
Absolutely, that has been on my mind for quite some time now. I've been discussing that with Andrei as well and it's clear that we need some proper abstraction over the types, hiding the protobufs. I think this is the first most important thing required for making Mir more useful as a stand-alone library. It will definitely happen as soon as there is some time for it.

pkg/checkpoint/checkpoint.go
adlrocha: Is not critical for spacenet, I am adding my own abstraction to make them CBOR-serializable here: https://github.com/consensus-shipyard/lotus/blob/b5b81dae0c451bf26c183d17795d38c38a460a37/chain/consensus/mir/types.go#L126-L190

It adds a `serde` overhead, but I think it is worth (and low) for now.

PR #282, opened 2022-10-24
Methods for stable checkpoint (de)serialization

matejpavlovic:


adlrocha (approved):
LGTM, this will come really handy. Thanks!

PR #284, opened 2022-10-26
Add test for reconnection and fix one more bug

dnkolegov:


pkg/net/libp2p/libp2p_transport_test.go
```diff
@@ -598,6 +598,51 @@ func TestLibp2p_OneConnectionInProgress(t *testing.T) {
...
+
+	initialNodes := m.Membership(nodeA)
+	initialNodes[nodeB] = bBadAddr
+	t.Log(initialNodes)
+
+	t.Log(">>> connecting to a failed node")
+	err = a.Start()
+	require.NoError(t, err)
+	a.Connect(initialNodes)
+	m.testEventuallyNotConnected(nodeA, nodeB)
```
matejpavlovic, line 627: Not sure `EventuallyNotConnected` is what we want here. As I understand it, this test trivially succeeds even if the nodes eventually connect, if it happens to poll the state before the connection is established.

matejpavlovic (commented):
Not sure the test is guaranteed to work... See comment.

It would also be nice to have a test that really tests sending messages. For example, periodically send a message to a connected node, then disconnect the node causing a few messages to be dropped, and then re-establish the connection and make sure the messages can get through agian.

dnkolegov:
> Not sure the test is guaranteed to work... See comment.
> 
> It would also be nice to have a test that really tests sending messages. For example, periodically send a message to a connected node, then disconnect the node causing a few messages to be dropped, and then re-establish the connection and make sure the messages can get through agian.

We have a such test!

https://github.com/filecoin-project/mir/blob/dd333c00d1f76bf5a1546b4d67bef3d8c16a390c/pkg/net/libp2p/libp2p_transport_test.go#L259

Let me know if that test does not correspond to your scenario.

pkg/net/libp2p/libp2p_transport_test.go
dnkolegov: The test works as follows: node A tries to connect to node B. node A is connecting to the close port of node B. We check that they are not connected (they never were connected). Then A is trying to connect to the open port of node B. And we check that nodes A succeeds.

matejpavlovic: What I wanted to say is the following: Calling `Connect` (that starts connecting in the background) and immediately testing whether the node is *not* connected creates a form of a race. Even if the node actually did connect one millisecond after you call `testEventuallyNotConnected`, the test would still succeed (an probably should fail)

matejpavlovic:
I'm not sure how exactly that test works...
```go
t.Log(">>> disconnecting nodes")
	m.disconnect(nodeA, nodeB)
	m.testEventuallyNotConnected(nodeA, nodeB)

	t.Log(">>> sending messages after disconnection")
	m.testEventuallySentMsg(nodeA, nodeB, &messagepb.Message{})
	m.testThatSenderIs(<-nodeBEventsChan, nodeA)
	m.testEventuallySentMsg(nodeA, nodeC, &messagepb.Message{})
	m.testThatSenderIs(<-nodeCEventsChan, nodeA)
```
You disconnect, then you try send a message, and expect to be delivered anyway (because the module should automatically re-connect)? If that's the case, I think it doesn't work all the time. If it's not the case, what exactly does this part of the test do?

pkg/net/libp2p/libp2p_transport_test.go
dnkolegov: I see. To fix that I explicitly specified parameters for `openStream`: no retries, 1 sec connecting timeout, and no delay.
`testNeverConnected` checks connection during 5 sec, `openStream` establishes a connection during 1 sec.

pkg/net/libp2p/libp2p_transport_test.go
```diff
@@ -623,10 +623,15 @@ func TestLibp2p_OpeningConnectionAfterFail(t *testing.T) {
 	t.Log(">>> connecting to a failed node")
 	err = a.Start()
 	require.NoError(t, err)
+	a.params.MaxRetries = 1
+	a.params.MaxConnectingTimeout = 1 * time.Second
+	a.params.MaxRetryTimeout = 0
 	a.Connect(initialNodes)
 	m.testNeverNotConnected(nodeA, nodeB)
```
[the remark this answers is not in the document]
dnkolegov: Misprint

pkg/net/libp2p/libp2p_transport_test.go
matejpavlovic: I think even then the problem persists. What if the test succeeds even before the first connection attempt? Since connection happens in the background, this is well possible...

dnkolegov:
> because the module should automatically re-connect

Correct.

> I think it doesn't work all the time.
What do you mean by this?

matejpavlovic:
I think it doesn't work all the time, because there is a possibility arbitrarily many messages being dropped before the connection is established again.

dnkolegov:
> It would also be nice to have a test that really tests sending messages. For example, periodically send a message to a connected node, then disconnect the node causing a few messages to be dropped, and then re-establish the connection and make sure the messages can get through agian.

@matejpavlovic  I added the test. Please check whether it satisfies your requirements.

pkg/net/libp2p/libp2p_transport_test.go
```diff
@@ -461,6 +463,120 @@ func TestLibp2p_SendingWithTwoNodes(t *testing.T) {
...
+		m.CloseHostAll()
+		m.testNoConnections()
+	}(ctx)
+
+	sent := 0
+	sentBeforeDisconnect := 0
+	received := 0
+	disconnect := make(chan struct{})
+
+	testTime := time.Duration(15)
```
matejpavlovic, line 521: It's a bit misleading to have `testTime` in 15 "units of duration". It would be more readable to me if that was already in seconds, without the rest of the code multiplying it by seconds each time.

matejpavlovic (approved):
👍

pkg/net/libp2p/libp2p_transport_test.go
dnkolegov: Fixed

PR #286, opened 2022-10-31
Pass whole StableCheckpoint to SMR RestoreState

matejpavlovic:


adlrocha (approved):
Great, I think this solves my concerns and needs :) I'll bump the new version in Lotus see if it fixes the bug. Thanks!

PR #301, opened 2022-11-10
Add checkpoint and certificate serialization

matejpavlovic:
The stable checkpoint can be now deterministically serialized and deserialized.
The certificate can be added to and removed from the checkpoint.

### DIFF
--- a/pkg/checkpoint/checkpoint.go
+++ b/pkg/checkpoint/checkpoint.go
@@ -3,7 +3,7 @@ package checkpoint
 import (
 	"fmt"
 
-	"google.golang.org/protobuf/proto"
+	"github.com/fxamacker/cbor/v2"
 
 	"github.com/filecoin-project/mir/pkg/clientprogress"
 	"github.com/filecoin-project/mir/pkg/crypto"
@@ -12,6 +12,7 @@ import (
 	"github.com/filecoin-project/mir/pkg/pb/commonpb"
 	"github.com/filecoin-project/mir/pkg/serializing"
 	t "github.com/filecoin-project/mir/pkg/types"
+	"github.com/filecoin-project/mir/pkg/util/maputil"
 )
 
 // StableCheckpoint represents a stable checkpoint.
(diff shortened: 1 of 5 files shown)


pkg/checkpoint/checkpoint.go
```diff
@@ -3,7 +3,7 @@ package checkpoint
 import (
 	"fmt"
 
-	"google.golang.org/protobuf/proto"
+	"github.com/fxamacker/cbor/v2"
```
adlrocha, line 6: I know that is still "half-baked" and it requires some type conversions to make it work, but why not using `github.com/whyrusleeping/cbor` which is the one used by all the Filecoin ecosystem?

adlrocha (approved):
Minor comment about the CBOR package, but otherwise LGTM. 

Out of a curiosity, why did you start the PR from your fork instead of using a branch of the `mir` repo directly? 

When this is merged I'll start upgrading the dependencies of Lotus and we can probably cut a new `0.1.1` release to include this change. Thanks!

pkg/checkpoint/checkpoint.go
matejpavlovic: Yes I saw you are using the other one, but I found it quite hard to use compared to the `fxamacker` one.

- I couldn't find proper documentation for the `whyrusleeping` one
- The interface seems to be very low-level. At least your code seems to explicitly deal with types of the serialized fields. The `fxamacker` version just directly serializes the struct.
- `whyrysleeping` is at least 5 years old (which might be a good sign for stability), but `fxamacker` is a much more recent.
- `fxamacker` has decent documentation and a very simple interface.

I'm actually wondering why the rest of the Filecoin ecosystem doesn't use the newer one :-D (But I'm ready to be educated :) )

matejpavlovic:
> Out of a curiosity, why did you start the PR from your fork instead of using a branch of the mir repo directly?

I always do it that way because I find it cleaner, not polluting the main public repo with extra branches. And we don't loose anything in terms of collaboration, since others can still push to the PR branch, even if it is in my repo.

PR #303, opened 2022-11-11
Explicitly describe that the function is insecure and intended for testing

dnkolegov:
Let's explicitly say that the function is not secure.


matejpavlovic (approved):
👍

PR #327, opened 2023-01-13
chore: Upgrade libp2p to remove btcd dependency

adlrocha:
Update `go-libp2p-core` and remove `btcd` as a dependency to avoid the incompatibility in Lotus produced by picking up a newer version of `btcd`.


adlrocha:
@matejpavlovic, this should be ready for review and merge. I can confirm that it works in Lotus.

matejpavlovic (approved):
Perfect, thanks a lot Alfonso, I would have never found this problem!

PR #328, opened 2023-01-13
More features around leader selection policy

matejpavlovic:
- Add leader selection policy to checkpoints
- Make the leader selection policy configurable
- Clean up and re-organize the code.


pkg/util/issutil/config.go
```diff
@@ -160,13 +165,14 @@ func DefaultParams(initialMembership map[t.NodeID]t.NodeAddress) *ModuleParams {
...
-		MsgBufCapacity:     32 * 1024 * 1024, // 32 MiB
-		RetainedEpochs:     1,
+		InitialMembership:     initialMembership,
+		ConfigOffset:          2,
+		SegmentLength:         4,
+		NumBuckets:            len(initialMembership),
+		RequestNAckTimeout:    16,
+		MsgBufCapacity:        32 * 1024 * 1024, // 32 MiB
+		RetainedEpochs:        1,
+		LeaderSelectionPolicy: Simple,
```
ranchalp: Why use Simple not Blacklist?

matejpavlovic: I made the `Simple` policy the default one to not break the interface, the `Simple` policy has implicitly been the default one before, as there was no other one. We can use `Blacklist` as the default in the future, but for now I'd prefer to have it "opt-in" rather than "opt-out".

PR #331, opened 2023-01-19
Verify signatures based on the membership

dnkolegov:
This PR changes the logic of how signatures are verified.
The current approach can lead to DoS attacks: it is enough to have one corrupted/incorrect signature, and the certificate will be alleged invalid. Another example is that the validator removed from the membership can add a signature to a certificate. In this case, since this validator is not contained in the membership but is contained in the certificate, the last one will also be invalid.


dnkolegov:
We found another way to implement a necessary mechanism for reconfiguration - https://github.com/filecoin-project/mir/pull/332

PR #337, opened 2023-01-24
New libp2p transport

matejpavlovic:
New implementation of the libp2p transport module with
- a non-blocking `Send()` method,
- concurrent sending to different nodes, and
- a bug fix related to closing incoming connections


pkg/net/libp2p/transport.go
```diff
@@ -0,0 +1,294 @@
...
+	tr.connectionsLock.RUnlock()
+
+	// Read values from the counter channel (aggregator) until enough successful
+	// connections (nil error) have been reached or an error occurs.
+	numConnected := 0
+	for numConnected < n {
+		select {
+		case err := <-counterChan:
+			if err != nil {
+				return // TODO: Return an error here (adaptation of the Transport interface needed)
```
dnkolegov, line 173: Should we address it outside of this PR? If not, let's create an issue to not forget about this

dnkolegov (approved):
LGTM!
Great job, Matej! Currently, after going through the code, I do not have any comments.
I added one minor note.

pkg/net/libp2p/transport.go
dnkolegov: Added https://github.com/filecoin-project/mir/issues/339

PR #346, opened 2023-02-13
Define persistent client interface

matejpavlovic:


pkg/client/interface.go
```diff
@@ -0,0 +1,36 @@
...
+
+// The Client represents an SMR client that produces new transactions (using NewTX)
+// that can be submitted to the SMR system.
+// It sequentially assigns transaction numbers to all created transactions.
+// The Client is informed (using Done)
+// when a transaction is considered applied (or otherwise safe to stop caring about).
+// All created transactions that have not been confirmed this way are considered pending.
+// The client is expected to persistently store all pending transactions,
+// in case they need to be resubmitted after recovering from a potential crash.
+type Client interface {
```
dnkolegov, line 16: Can we rename the interface or package to something that would describe the behavior?
Otherwise, we will have `client.Client` in the code. Even `smr.Client` sounds better. Is that correct?

pkg/client/interface.go
```diff
@@ -0,0 +1,36 @@
...
+
+	// NewTX creates and returns a new transaction of type txType, containing data.
+	// The returned transaction will be assigned the next transaction number in sequence,
+	// corresponding to the number of transactions previously created by this client.
+	// Until Done is called with the returned transaction's number,
+	// the transaction will be pending, i.e., among the transactions returned by Pending.
+	// The transaction need not necessarily yet be written to persistent storage when NewTX returns.
+	// Use the separate Sync() method to guarantee persistence.
+	NewTX(txType uint64, data []byte) *requestpb.Request
```
dnkolegov, line 25: All methods do not return errors. Is that correct keeping in mind that clients can be persistent and store data in persistent storage? Interacting with such storage may trigger errors.

dnkolegov (approved):

pkg/client/interface.go
matejpavlovic: I'd like it to be a separate package though, specific to the client, and both names (package and type) should be descriptive. The standard library does that too (`time.Time`). I wouldn't say it's a big deal, so I'd suggest leaving it, unless you have strong objections...

pkg/client/interface.go
matejpavlovic: Very good point!
Fixed.
I was hesitating whether to add an error return value to `Pending`, but I put it there too. I can imagine implementations that could return an error there as well.

matejpavlovic: Very good point!
Fixed.
I was hesitating whether to add an error return value to `Pending`, but I put it there too. I can imagine implementations that could return an error there as well.

dnkolegov (approved):
LGTM!

PR #347, opened 2023-02-13
libp2p mesh tests

dnkolegov:
This PR adds a test that simulates messaging in a network with ten nodes and adding five new nodes.

### DIFF
--- a/pkg/net/libp2p/transport_test.go
+++ b/pkg/net/libp2p/transport_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math/rand"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -835,7 +836,7 @@ func TestMeshMessaging(t *testing.T) {
 
 	senders := sync.WaitGroup{}
 
-	var received, sent int
+	var received, sent int64
 
 	for i := 0; i < N; i++ {
 		for j := 0; j < N; j++ {
@@ -857,7 +858,7 @@ func TestMeshMessaging(t *testing.T) {
 						if err != nil {
 							t.Logf("%v->%v sending: %v", src, dst, err)
 						} else {
-							sent++
+							atomic.AddInt64(&sent, 1)
 						}
 
 					}
@@ -883,7 +884,7 @@ func TestMeshMessaging(t *testing.T) {
 					}
 					_, valid := e.Iterator().Next().Type.(*eventpb.Event_MessageReceived)
 					require.Equal(m.t, true, valid)
-					received++
+					atomic.AddInt64(&received, 1)
 				}
 			}
 		}(ch.ownID, ch.incomingMessages, ch.stop)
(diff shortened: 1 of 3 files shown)


pkg/net/libp2p/transport_test.go
```diff
@@ -770,3 +825,121 @@ func TestLibp2p_OpeningConnectionAfterFail(t *testing.T) {
...
+
+	t.Logf(">>> connecting nodes")
+	for i := 0; i < N; i++ {
+		m.transports[nodes[i]].Connect(initialNodes)
+	}
+
+	for i := 0; i < N; i++ {
+		for j := 0; j < N; j++ {
+			if i == j {
+				continue
```
matejpavlovic, line 855: This could be simplified by inverting the condition (`i != j`), putting the next line inside, and removing the `continue` altogether.
This applies to other places in the code too.

pkg/net/libp2p/transport_test.go
```diff
@@ -770,3 +825,121 @@ func TestLibp2p_OpeningConnectionAfterFail(t *testing.T) {
...
+	t.Log(">>> sending messages")
+
+	var received, sent int64
+	var senders, receivers sync.WaitGroup
+
+	sender := func(src, dst types.NodeID) {
+		defer senders.Done()
+
+		for i := 0; i < 100; i++ {
+			time.Sleep(time.Duration(rand.Intn(200)) * time.Millisecond) // nolint
```
matejpavlovic, line 870: This looks like a long sleep in total (10 seconds on expectation). Do we need to wait this long or can we make the random delay shorter?

pkg/net/libp2p/transport_test.go
```diff
@@ -770,3 +825,121 @@ func TestLibp2p_OpeningConnectionAfterFail(t *testing.T) {
...
+			err := m.transports[src].Send(dst, &messagepb.Message{DestModule: "iss", Type: &messagepb.Message_Iss{}})
+			if err != nil {
+				t.Logf("%v->%v failed to send: %v", src, dst, err)
+			}
+			atomic.AddInt64(&sent, 1)
```
matejpavlovic, line 875: Should we increment the counter even if the sending failed?

matejpavlovic (commented):
Thanks Denis! Left just a few minor comments.

I'm not quite sure though I understood exactly what the test is doing. It connects NxN nodes, starts sending messages between all nodes (connected and unconnected) and then connects the rest?
Could you pls write a few comments to explain?

pkg/net/libp2p/transport_test.go
dnkolegov: Fixed.

pkg/net/libp2p/transport_test.go
dnkolegov: We definitely can!

pkg/net/libp2p/transport_test.go
dnkolegov: Fixed

matejpavlovic (approved):
Thanks Denis! Let's merge :-)

PR #348, opened 2023-02-17
Mir audit: Batchfetcher and mempool modules update to DSL codegen.

ranchalp:


ranchalp:
Builds upon #336

ranchalp:
Added a fix for #353 , the fix may interfere with `mpdsl.UponRequestTransactions` if that call is for a number of txIDs that are not in memory anymore, but I could not find a single event being created for this purpose.

ranchalp:
Also, found #352 along the way. We should discuss how much of an issue we think that could be (specially with the latest's availability module #336 .

ranchalp:
Forced push to account for possibility of multiple client's requests with the same txID when fixing #353.

ranchalp:
Added support for map types to codegen, although it is not being used at the moment. The mir.type field must define both key and val types in the following way `keytype|valtype`

matejpavlovic:
I'm putting this on hold until #336 is polished and merged, let's have a look at this one afterwards.

pkg/batchfetcher/batchfetcher.go
```diff
@@ -104,16 +106,12 @@ func NewModule(mc *ModuleConfig, epochNr t.EpochNr, clientProgress *clientprogre
 
 	// The AppRestoreState handler restores the batch fetcher's state from a checkpoint
 	// and forwards the event to the application, so it can restore its state too.
-	eventpbdsl.UponAppRestoreState(m, func(restoreState *eventpb.AppRestoreState) error {
-
-		// Load the checkpoint from the received event.
-		chkp := checkpoint.StableCheckpointFromPb(restoreState.Checkpoint)
```
matejpavlovic, line 113: I'd actually prefer using the checkpoint abstraction. It might be marginally les performant, but I find it cleaner.

pkg/batchfetcher/batchfetcher.go
```diff
@@ -104,16 +106,12 @@ func NewModule(mc *ModuleConfig, epochNr t.EpochNr, clientProgress *clientprogre
...
-
+	eventpbdsl.UponAppRestoreState(m, func(chkp *checkpointpb.StableCheckpoint) error {
 		// Update current epoch number.
-		epochNr = chkp.Epoch()
+		epochNr = t.EpochNr(chkp.Snapshot.EpochData.EpochConfig.EpochNr)
```
matejpavlovic, line 112: If we keep using the checkpoint abstraction, we don't need to rely on the internal format of the data structure (in case it ever changes). And the code will also be easier to look at.

codegen/model/types/map.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"reflect"
+
+	"github.com/dave/jennifer/jen"
+)
+
+var _thisPackage = reflect.TypeOf(Map{}).PkgPath()
+
+// Slice is used to represent repeated fields.
```
matejpavlovic, line 11: Forgot copy-pasted comment?

matejpavlovic (commented):
Great job! Nicely structured PR.
Even better, adding the map support to the codegen could have been done in a separate PR to make it easy to review separately. In fact, even the updates to the batch fetcher and mempool could go to different PRs next time. Still, very nice job!!

Let's try to get @xosmig 's review on the map support in the code generation, I'm not too familiar with the code and it will probably take him 5 minutes.

Andrei, would you have time to look at the codegen part? It's just 2 files and probably it won't even take you 5 minutes. Thanks a lot!

codegen/model/types/parser.go
```diff
@@ -238,15 +238,26 @@ func (p *Parser) parseField(
 	}, nil
 }
 
-func (p *Parser) getFieldType(goType reflect.Type, protoField protoreflect.FieldDescriptor) (Type, error) {
+func (p *Parser) getFieldType(goType reflect.Type, protoField protoreflect.FieldDescriptor, mapType bool) (Type, error) {
 	// TODO: Since maps are not currently used, I didn't bother supporting them yet.
+	//if goType.Kind() == reflect.Map {
+	//	return nil, fmt.Errorf("map fields are not supported yet")
+	//}
```
xosmig, line 245: Probably can be removed now :)

xosmig (approved):
I added a few notes regarding the type annotation syntax, but other than that, the codegen part looks great, thank you! :)

codegen/model/types/parser.go
```diff
@@ -263,6 +274,15 @@ func (p *Parser) getFieldType(goType reflect.Type, protoField protoreflect.Field
...
 
 	if mirTypeOption != "" {
+		sepMapIdx := strings.Index(mirTypeOption, "|")
+		if sepMapIdx != -1 { // it is a map type
+			if !mapType { //key value
+				mirTypeOption = mirTypeOption[:sepMapIdx]
+			} else { // map value
+				mirTypeOption = mirTypeOption[sepMapIdx+1:]
+			}
+		}
```
xosmig, line 284: Also, I would be curious to know your opinion on potential alternative syntaxes for the annotation.

A few natural examples (in no specific order):
1. using -> as the separator and trimming the whitespaces around it
2. using map<keyType, valueType> to follow the syntax of protobufs
3. using map[keyType]valueType to follow the syntax of Go
4. using two separate annotations (mir.key_type) and (mir.value_type) and throwing an error if (mir.type) is used for a map type

One potential concern about all options except the last one (including the current version) is the length of the annotation, since, as of now, the type annotation has to include the full package path for each type.

xosmig: Option (4) could look somewhat like this:

```protobuf
message MapMessage {
    map<string, int> mapValue [
        (mir.key_type) = "github.com/awesome-company/awesome-project/package.KeyType",
        (mir.value_type) = "github.com/awesome-company/awesome-project/package.ValueType",
    ];
}
```

The current version may be a bit hard to parse visually:

```protobuf
message MapMessage {
    map<string, int> mapValue [(mir.type) = "github.com/awesome-company/awesome-project/package.KeyType|github.com/awesome-company/awesome-project/package.ValueType"];
}
```

It could probably be helped to a certain extent by trimming the whitespaces around the separator.

matejpavlovic:
Thanks a lot @xosmig for the comments, it's very helpful!

codegen/model/types/parser.go
matejpavlovic: Intuitively, option (4) seems the nicest to me. On top of what you already said, it seems to avoid the necessity of string parsing magic. @ranchalp what do you think?

pkg/batchfetcher/batchfetcher.go
```diff
@@ -53,14 +53,14 @@ func NewModule(mc *ModuleConfig, epochNr t.EpochNr, clientProgress *clientprogre
...
-	dsl.UponEvent[*eventpb.Event_DeliverCert](m, func(cert *eventpb.DeliverCert) error {
+	eventpbdsl.UponDeliverCert(m, func(sn t.SeqNr, cert *apbtypes.Cert) error {
 		// Create an empty output item and enqueue it immediately.
 		// Actual output will be delayed until the transactions have been received.
 		// This is necessary to preserve the order of incoming and outgoing events.
 		item := outputItem{event: nil}
 		output.Enqueue(&item)
 
-		if cert.Cert.Type == nil {
+		if cert.Pb().Type == nil {
```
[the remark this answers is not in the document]
ranchalp: This is something I have been thinking about it myself. I will add the check as it is a (small) optimization, but I think we should open a TODO/issue for a more standard way to check for empty certificates (in the view change it is a `[]byte{}`, in the availability module it is `[]byte{0}` and here it seems to be a nil. This is not immediate though as it might even make sense to have different empty certs depending on which module should check.

pkg/batchfetcher/batchfetcher.go
ranchalp: ditto, need to come back to this to tag the type on protobufs and thus removing `t.EpochNr` once we are back to using the codegen code.

pkg/batchfetcher/batchfetcher.go
ranchalp: Done as requested, but may come back to this once I pass through the checkpoint module. Bear in mind that `checkpoint.` refers to a manually written package, whereas `checkpointpbtypes.` is automatically generated. I believe minimizing the amount of manually written code and maximizing tags in protobufs is the way to go. If types change later at least we know that they must be changed just in protobufs.

pkg/batchfetcher/batchfetcher.go
```diff
@@ -104,16 +106,12 @@ func NewModule(mc *ModuleConfig, epochNr t.EpochNr, clientProgress *clientprogre
...
-		chkp := checkpoint.StableCheckpointFromPb(restoreState.Checkpoint)
-
+	eventpbdsl.UponAppRestoreState(m, func(chkp *checkpointpb.StableCheckpoint) error {
 		// Update current epoch number.
-		epochNr = chkp.Epoch()
+		epochNr = t.EpochNr(chkp.Snapshot.EpochData.EpochConfig.EpochNr)
 
 		// Load client progress.
-		clientProgress.LoadPb(chkp.StateSnapshot().EpochData.ClientProgress)
+		clientProgress.LoadPb(chkp.Snapshot.EpochData.ClientProgress)
```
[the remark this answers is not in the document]
ranchalp: ditto.

pkg/batchfetcher/events/events.go
```diff
@@ -8,6 +8,7 @@ import (
 	t "github.com/filecoin-project/mir/pkg/types"
 )
 
+// These are specific events that are not automatically generated but are necessary for the batchfetcher to queue them and tream them in order
```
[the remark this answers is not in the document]
ranchalp: yes, done, "treat".

pkg/mempool/simplemempool/internal/parts/formbatches/formbatches.go
```diff
@@ -26,6 +26,7 @@ func IncludeBatchCreation(
 		State:    commonState,
 		NewTxIDs: nil,
 	}
+	txIDCount := make(map[string]int, 0)
```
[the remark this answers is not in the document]
ranchalp: Yes that is the reason. I am aware that with the current implementation the app logic is responsible for filtering collisions and just keeping one copy, but here the purpose is to prevent the state from blowing app by keeping in memory all transactions (i.e. removing after decision). The fix is clear. However, if I remove a transaction after the first decision and that transaction is proposed again (by a malicious proposer), then, after deciding on this proposal, line `txIDs = append(txIDs, txID)` will throw a runtime error, because `tx := state.TxByID[string(txID)]` will return nil (and `append(slice, nil)` throws a runtime error.

codegen/model/types/map.go
ranchalp: Correct, thanks, fixed.

protos/checkpointpb/checkpointpb.proto
```diff
@@ -21,15 +21,19 @@ message Checkpoint {
...
+
     commonpb.EpochConfig epoch_config      = 1;
     StableCheckpoint     stable_checkpoint = 2;
     EpochProgress        epoch_progress    = 3;
   }
 }
 
 message StableCheckpoint {
-  //option (mir.event) = true; TODO add support for maps from dsl's codegen?
+  option (mir.event) = true; //TODO add support for maps from dsl's codegen?
```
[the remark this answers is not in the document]
ranchalp: It is, but I clarified further, see above explanation.

codegen/model/types/parser.go
```diff
@@ -238,15 +238,26 @@ func (p *Parser) parseField(
 	}, nil
 }
 
-func (p *Parser) getFieldType(goType reflect.Type, protoField protoreflect.FieldDescriptor) (Type, error) {
+func (p *Parser) getFieldType(goType reflect.Type, protoField protoreflect.FieldDescriptor, mapType bool) (Type, error) {
```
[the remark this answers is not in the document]
ranchalp: Done (as indicated below)

codegen/model/types/parser.go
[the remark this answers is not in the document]
ranchalp: Done

codegen/model/types/parser.go
ranchalp: Done, the new code also supports defining only the key (by either not writing '|' or writing it as the last character when tagging).

ranchalp: Done.

codegen/model/types/parser.go
ranchalp: Option 4 sounds good. Just did it, and took the extra effort to also support defining only the key or only the value types (which is related to other comments of this PR and will enable for removing even more boilerplate code).

pkg/batchfetcher/batchfetcher.go
ranchalp: Update: after addressing @xosmig 's comments I decided to also add support for map types defining only the key or the value type. So in a new upcoming PR we can clean up this code further. Not doing it on this PR for the purpose of not overcomplicating it at this stage.

codegen/model/types/parser.go
ranchalp: I also left as TODO in issue #356 the last bit that I think we need to completely convert all proto messages with the codegen. Will be doing that in an upcoming PR. Thanks for your comments! :)

ranchalp:
Thanks @xosmig and @matejpavlovic for the comments! :)

pkg/mempool/simplemempool/internal/parts/formbatches/formbatches.go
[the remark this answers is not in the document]
matejpavlovic: I think there are 2 concepts being mixed here: (1) proposing transactions and (2) adding transactions to the mempool.

Unless I missed something (possible), the mempool is a purely local data structure that serves as a buffer for transactions waiting to be proposed by _this node_, regardless of other malicious nodes are proposing. Or does the mempool get involved in any way during other nodes' proposals? It is possible, though, that some client (not even necessarily a malicious one) can send the same transaction to the same node more than once.

But this should be addressed differently in my opinion. It would be both more natural and way simpler to just keep at most one copy of a transaction in the mempool.
1. Remove the counter
2. When adding transactions, ignore those that are already inside according to the index (note that this is easy, as everything is executed sequentially)

With this approach, is there still a problem that the counter would solve?

codegen/model/types/map.go
matejpavlovic: Is it? 😄

pkg/batchfetcher/batchfetcher.go
```diff
@@ -122,7 +126,7 @@ func NewModule(mc *ModuleConfig, epochNr t.EpochNr, clientProgress *clientprogre
 		// Forward the RestoreState event to the application.
 		// We can output it directly without passing through the queue,
 		// since we've just reset it and know this would be its first and only item.
-		eventpbdsl.AppRestoreState(m, mc.Destination, chkp)
+		eventpbdsl.AppRestoreState(m, mc.Destination, checkpointpbtypes.StableCheckpointFromPb(chkp.Pb()))
```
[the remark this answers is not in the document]
ranchalp: You are right, this is not necessary anymore, removing...

pkg/mempool/simplemempool/internal/parts/formbatches/formbatches.go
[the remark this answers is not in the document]
ranchalp: Actually yes you are absolutely right, I thought the mempool was storing transactions received from batches by the availability module, which in this case is of course the responsibility of the fakebatchdb (and obviously the fakebatchdb is already known to blow up the state by keeping the database in plain maps in memory). I just updated the code to remove the counter on the txID (and assume as elsewhere that the app logic prevents collisions).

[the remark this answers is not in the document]
matejpavlovic: What does need to be fixed, though, is checking for the presence of the transaction when adding it, and only add it if it is not yet in the `TxByID` index (otherwise ignore it).

codegen/model/types/parser.go
xosmig: Sorry, I just realized that there is actually a perfectly valid use-case for using all three of `(mir.value_type)`, `(mir.key_type)`, and `(mir.type)`. Namely, when there is a wrapped map type.

Something like:
```go
// Membership represents the list of participants with their respective weights.
type Membership map[NodeID]int
```

xosmig: ah, but maybe it's already supported by the current code :)

pkg/mempool/simplemempool/internal/parts/formbatches/formbatches.go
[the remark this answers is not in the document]
ranchalp: done

codegen/model/types/parser.go
xosmig: No, I guess it's not, but it should be the matter of removing the extra check

codegen/model/types/parser.go
```diff
@@ -262,6 +271,18 @@ func (p *Parser) getFieldType(goType reflect.Type, protoField protoreflect.Field
 		return Error{}, nil
 	}
 
+	if ps != defaultParserState {
+		if mirTypeOption != "" {
+			return nil, fmt.Errorf("mir.type option cannot be used on maps")
+		}
+
```
xosmig, line 278:
```suggestion
```

codegen/model/types/parser.go
xosmig: Hm, there is an issue though: it won't be consistent with how `(mir.type)` operates on slices

xosmig: maybe, in the long term (perhaps, not in this PR), it would be better to migrate slices to using a pair of annotations:
1. `(mir.value_type)` for the underlying type
2. `(mir.type)` for wrapping the slice itself

matejpavlovic: Could that be somehow related to a solution to #356 ?

But I agree that this should definitely be addressed in a different PR.

Issue #335, opened 2023-01-24
Orderer's preprepare not verifying certificate


ranchalp:
The current Orderer's treatment of a preprepare msg sent by the primary does not verify the certificate. A malicious proposer could propose a bogus certificate and have it decided.

PR #361, opened 2023-03-01
Fix to #335

ranchalp:
Fix to #335 after decision (not to wait for timeout) and after rebasing with latest PRs.

### DIFF
--- a/pkg/batchfetcher/batchfetcher.go
+++ b/pkg/batchfetcher/batchfetcher.go
@@ -3,11 +3,11 @@ package batchfetcher
 import (
 	"fmt"
 
+	bfevents "github.com/filecoin-project/mir/pkg/batchfetcher/events"
 	"github.com/filecoin-project/mir/pkg/checkpoint"
 	"github.com/filecoin-project/mir/pkg/logging"
 	checkpointpbtypes "github.com/filecoin-project/mir/pkg/pb/checkpointpb/types"
 
-	bfevents "github.com/filecoin-project/mir/pkg/batchfetcher/events"
 	availabilitypbdsl "github.com/filecoin-project/mir/pkg/pb/availabilitypb/dsl"
 	apbtypes "github.com/filecoin-project/mir/pkg/pb/availabilitypb/types"
 	bfeventstypes "github.com/filecoin-project/mir/pkg/pb/batchfetcherpb/events"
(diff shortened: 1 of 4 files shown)


ranchalp:
Again lint works for me locally, still have to figure out what it is (probs the version being used)

matejpavlovic:
It seems that the certificates are being checked in the PBFT orderer, which implies that the orderer needs to know that it is ordering certificates. I feel like this specializes the orderer too much. It already wouldn't apply for the checkpoints certificates, for which the orderer is now also used. The result is really "hacky" code that checks if an availability module has been defined.

Since the certificates are checked only after ordering anyway now, I'd strongly prefer doing it directly from `iss.go`, after the availability certificate is delivered from the SB instance and distinguished from a checkpoint. I.e., right before inserting it into the `commitLog`. If the check succeeds, the delivered certificate is inserted, if not, an empty one is inserted instead.

matejpavlovic:
And for the linting, you can see the offending lines directly in the code even here in GitHub, when you click on "Files Changed" on the top of this conversation.

ranchalp:
Done, though I do think that it is also not idiomatic that the orderer is the one that communicates with the availability module and receives the certificate, but it is later ISS who communicates with the availability module to request a verification of that certificate sent to the orderer. 
I think it would be cleaner to have a specific availability module that simply returns `verified` for the checkpoint ordering, instead of having on one side the ValidityChecker (only really used by checkpoint ordering) and the availability module checking (only used by non-checkpoint ordering). 
Perhaps we would need to lose the optimization of waiting for the timeout if the certificate is invalid (if checkpoint ordering needs to verify validity on start, which I do not think it does?).
What do you think? Either way, it is done as requested already :)

matejpavlovic:
Yes I see the point, and I was thinking about it when implementing the checkpoint verification. The issue is that the checkpoint verification is synchronous and doesn't need to issue any events, while the availability certificate verification does (it needs to communicate with the availability module).

A clean solution would be to have a separate module abstraction with a unified interface for verifying proposals (for availability certificates, the availability module would implement that interface). I wanted to do that eventually, but it hasn't been a priority until now (and still isn't). But let's create an issue for that and address it later.

ranchalp:
Sounds good! merging then.

PR #374, opened 2023-03-23
Stubborn message sending in libp2p transport

matejpavlovic:
Previously, if an error occurred while sending a message in the libp2p transport, the module dropped the message and tried to reconnect. If messages were very sparse and the other side of the connection gave up receiving after some time after each incoming connection, all messages would get lost.

Now, if the libp2p transport fails to send a message, it does not drop it, but instead tries to reconnect and, if reconnection succeeds, immediately re-tries to send the same message until it is sent successfully.

### DIFF
--- a/pkg/net/libp2p/remoteconnection.go
+++ b/pkg/net/libp2p/remoteconnection.go
@@ -241,6 +241,16 @@ func (conn *remoteConnection) process() {
 		default:
 		}
 
+		// Create a network connection if there is none.
+		if conn.stream == nil {
+			if err := conn.connect(); err != nil {
+				// Unless the connection is closing, connect() will keep retrying to connect indefinitely.
+				// Thus, if it returns an error, it means that there is no point in continuing the processing.
+				conn.logger.Log(logging.LevelWarn, "Gave up connecting", "err", err)
+				return
+			}
+		}
+
 		// Get the next message and encode it if there is no pending unsent message.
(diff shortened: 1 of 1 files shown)


pkg/net/libp2p/remoteconnection.go
```diff
@@ -224,9 +224,38 @@ func (conn *remoteConnection) process() {
...
+			return
+		default:
+		}
+
+		// Get the next message and encode it if there is no pending unsent message.
+		if msgData == nil {
+			select {
+			case <-conn.stop:
+				return
+			case msg := <-conn.msgBuffer:
```
dnkolegov, line 249: Tests do not work because execution is blocked in this select and waits to receive a stop signal or msg.
The API contains `Connect`; because of that, as a user, I suggest that if I call `Connect` the peers are connected.
But you probably implicitly changed API by changing the implementation and now to connect a user must send a message.

dnkolegov (changes requested):

pkg/net/libp2p/remoteconnection.go
matejpavlovic: That's right, thanks!!
I've just fixed that. Now the underlying connections are established right after `Connect` is called (even though this usually doesn't finish before `Connect` returns).

dnkolegov (approved):

PR #382, opened 2023-04-04
Fix transaction deduplication bug in batch fetcher

matejpavlovic:
If the batch fetcher received transactions from the availability module in a different order than the order in which they should be delivered, it could happen that the deduplication code removed the transactions from the wrong batches.
This is now fixed by performing deduplication immediately before delivery instead of upon the reception of transaction payloads from the availability module.


ranchalp:
LGTM! Good catch by the way!

PR #395, opened 2023-04-09
Verify senders and signers are part of membership 

### LINKED ISSUE #384: PBFT View Change does not verify senders are part of current configuration
The View Change messages sent by both the primary and other nodes are sent to the crypto module for verification, but whether the signers/sender are part of the current configuration is not verified neither by the view change nor by the crypto module. The crypto module implements a function to delete keys from the set of valid keys for verification, but this function does not seem to be used anywhere. This means that an old replica could forge view change messages and cause disagreements.

Note that this is an observation made after reviewing the code for PBFT's view change, and not 100% sure if perhaps another module (for example, the transport module) restricts received messages to the current configuration. But the transferable authentication of the view change messages sent by the primary to the rest of the nodes can definitely be forged (they are all sent from the primary and neither the orderers nor the crypto module verify the current configuration on its signers).

Two options to resolve this: 
 a) Start deleting public keys from old configurations. We should take care with this in case there are some messages that we should take into consideration. This is in my opinion dirtier because of this concern.
 b) Verify the sender and all the signatures are part of the current configuration. This should be fairly easy, and mostly as simple as verifying this at the beggining of `applyMsgReceived` (or at least as easy as that except for specific messages that we need to allow old-members or someone else to send/sing). On top of that, check on instances that use transferable authentication (the `newView` message only at the moment AFAIK).

ranchalp:
Fixes part of #384 (but sill work to do there, do not close issue until fixed for ISS too):

- [x] availability module: all messages and transferred authentication of signatures of certificates (check that sender of message and all signers of certificate are members)
- [x] checkpoint module: all messages and transferred authentication of signatures of certificates (check that sender of message and all signers of certificate are members)
- [x] orderer's module: all messages and transferred authentication of signatures of certificates (check that sender of message and all signers of certificate are members)

### DIFF
--- a/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go
+++ b/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go
@@ -2,6 +2,7 @@ package batchreconstruction
 
 import (
 	"fmt"
+
 	"github.com/filecoin-project/mir/pkg/logging"
 	"github.com/filecoin-project/mir/pkg/util/sliceutil"
 
--- a/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go
+++ b/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go
@@ -2,6 +2,8 @@ package batchreconstruction
 
 import (
 	"fmt"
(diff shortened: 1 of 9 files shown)


matejpavlovic (approved):
Nice and readable, very easy to review.
Great job, thanks!

PR #414, opened 2023-04-20
multisigcollector: Fix nil deref

abread:
During testing I noticed this bug. When processing the NodeSigsVerified event, if they are not `allOk` the request is replied to, deleted, but the handler does not return, leading to a nil deref when acessing the now-deleted request.


matejpavlovic (approved):
Good catch, thanks a lot!

Issue #356, opened 2023-02-23
Tagging `bytes` protobuffers with mir type results in wrong conversion


ranchalp:
Protobuffers `bytes` are converted to slices `[]uint8` in Go. Because of this, when tagging a `bytes` type with a go type using `mir.type`  (for example, `type TxID []byte`) in protobuffers, the codegen actually takes the tagged type and uses it to replace the underlying type of the slice, i.e. `uint8`. As a result, the resulting type is `[]TxID` instead of `TxID`. A similar behavior is expected when tagging map types with `mir.key_type` and `mir.value_type`.

The proposal consists of creating a new tag, `mir.outer_type`, that if defined tells the codegen to actually convert the entire the entire slice or map, instead of the underlying types.

xosmig:
I imagine, with the suggested fix (a new annotation), there could be an issue with `repeated bytes` fields.

xosmig:
Although, the feature of wrapping the repeated field, as well as maps, would be nice on its own :)
(either with a new annotation or with `(mir.type)` as suggested in https://github.com/filecoin-project/mir/pull/348#discussion_r1121774305)

Issue #360, opened 2023-03-01
Clean up empty decision/certificate 


ranchalp:
At the moment, at different modules and parts of the codes there are different standards to check for an empty decision/certificate. Either `[]byte{0}`, `[]byte{},` `nil` or even perhaps `.Aborted` to refer to empty decisions. It is not immediate that we should have one to rule them all, as different modules may conceptually refer to different empty certificates. The issue requires some time to conceptually reevaluate the current implementation and implement changes to make empty certificates consistent and self-explanatory.

PR #416, opened 2023-04-22
Simplify multisig collector

matejpavlovic:
First contribution to #360.
The main simplification is that empty MSC sub-certificates receive no special treatment now.
Since an availability certificate consists of a list of MSC sub-certificates, the natural way of representing an empty availability certificate (i.e., one that certifies the availability of no transactions) is to simply have zero sub-certificates in a certificates (and thus, naturally, it is automatically valid by the semantics of the "for all" quantifier, in the sense that no sub-certificates are invalid). Another big advantage is that the mempool does not need to be involved in the dirty business of using a special value for the batch ID.

To preempt the question why I removed [triggering of the `ComputeBatch` event upon reception of an empty batch when requests were pending](https://github.com/filecoin-project/mir/blob/1872c987bd6f6ad4a3648a71691c5ff8ffd363b4/pkg/availability/multisigcollector/parts/certcreation/certcreation.go#L102): Except in a very unlikely case where a CertRequest arrives immediately responding to another request with an empty certificate, or where new transactions arrive at the mempool concurrently with sending the empty certificate, this `ComputeBatch` event has no effect, as it will very quickly go to the mempool again, find nothing there, trigger the `UponNewBatch` again (with an empty batch), and find no more requests (as all of them just have been sent an empty certificate), and stops. So (except for the concurrency-related corner case) the old approach is equivalent to just never triggering `ComputeBatch` on an empty batch.
One inconvenience and the remedy is described in comments, but that's for another PR.


ranchalp:
> To preempt the question why I removed [triggering of the ComputeBatch event upon reception of an empty batch when requests were pending](https://github.com/filecoin-project/mir/blob/1872c987bd6f6ad4a3648a71691c5ff8ffd363b4/pkg/availability/multisigcollector/parts/certcreation/certcreation.go#L102): Except in a very unlikely case where a CertRequest arrives immediately responding to another request with an empty certificate, or where new transactions arrive at the mempool concurrently with sending the empty certificate, this ComputeBatch event has no effect, as it will very quickly go to the mempool again, find nothing there, trigger the UponNewBatch again (with an empty batch), and find no more requests (as all of them just have been sent an empty certificate), and stops. So (except for the concurrency-related corner case) the old approach is equivalent to just never triggering ComputeBatch on an empty batch.
One inconvenience and the remedy is described in comments, but that's for another PR.

This is true, but the loop will continue until the mempool provides a batch (i.e. the MSC will asynchronously generate a certificate as soon as there is a batch available). I do not see a problem with the infinite loops of events (that do not bloat memory) until there is a batch that is not empty, but I may be missing something.

pkg/mempool/simplemempool/parts/computeids/computeids.go
```diff
@@ -43,11 +42,7 @@ func IncludeComputationOfTransactionAndBatchIDs(
 
 	mppbdsl.UponRequestBatchID(m, func(txIDs []tt.TxID, origin *mppbtypes.RequestBatchIDOrigin) error {
 		data := make([][]byte, len(txIDs))
-		copy(data, txIDs)
-
-		if len(txIDs) == 0 {
-			mppbdsl.BatchIDResponse(m, origin.Module, emptybatchid.EmptyBatchID(), origin)
-		}
+		copy(data, txIDs) // TODO: Why do we copy the txIDs here? Is it because the hasher, technically, doesn't promise be read-only?
```
ranchalp, line 45: Not actually needed at the moment because of the type aliases so they are both the same type really. Once we fix #356 and define types (instead of using aliases), then the copy would make sense. Can be removed for now.

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
```diff
@@ -134,8 +134,8 @@ func IncludeCreatingCertificates(
 			return nil
 		}
 
-		if !certificate.receivedSig[from] {
-			certificate.receivedSig[from] = true
+		if !certificate.ReceivedSig[from] {
+			certificate.ReceivedSig[from] = true
```
ranchalp, line 138: Are these public struct fields being used anywhere outside the package?

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
```diff
@@ -95,17 +94,18 @@ func IncludeCreatingCertificates(
...
+			// If the batch is empty, immediately stop it and respond to all pending requests with an empty certificate.
 			delete(state.Certificates, context.reqID)
-			if len(state.RequestStates) > 0 { // do not do work for the empty batch if there is not even a request to answer
-				sendIfReady(m, &state, params, true)
-				if len(state.Certificates) == 0 {
-					apbdsl.ComputeCert(m, mc.Self)
-				}
-			}
-			return nil
+			sendEmptyCert(m, &state)
```
ranchalp: This seems to send an empty certificate when receiving a batch, while `sendIfReady` responds with all the so-far delivered batches if any available, and only with an empty cert if none are available. I may be missing something perhaps thinking

ranchalp (commented):
Great job! some questions remain :)

pkg/availability/multisigcollector/parts/batchreconstruction/batchreconstruction.go
```diff
@@ -55,39 +53,32 @@ func IncludeBatchReconstruction(
...
+			ReqOrigin:     origin,
+		}
 
-			}
-		} else {
-			// give empty batch and delete request
-			apbdsl.ProvideTransactions(m, origin.Module, []*requestpbtypes.Request{}, origin)
-			delete(state.RequestState, reqID)
+		for _, c := range certs {
+			batchIDString := msctypes.BatchIDString(c.BatchId)
```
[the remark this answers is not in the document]
ranchalp: I think my above comment may be deprecated after seeing #417 , but leaving it for completeness and for you to double check this important bit.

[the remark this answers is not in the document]
matejpavlovic: Yes, if `certs[i]` is nil, this would crash, but `certs[i]` is never nil, even irrespectively of #417. Already in this version (in fact, even before this PR), correct code never produces a nil entry in this list (not even ISS or the batch fetcher). What could happen, though, that a malicious node proposes a malformed certificate with a `nil` entry (once `nil` entries are supported - until then, we would crash even before getting here). This is a matter of input sanitization though, there already is an issue for that.

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
```diff
@@ -95,17 +94,18 @@ func IncludeCreatingCertificates(
...
-				}
-			}
-			return nil
+			respond(m, &state, []*mscpbtypes.Cert{})
+			// Note that a creation of another certificate is NOT started here.
+			// It will only be triggered by another certificate request (UponRequestCert).
+			// This creates a slight inconvenience that, if there is an empty batch no certificates will be pre-computed
+			// for the next request, and the next request will have to wait until the availability protocol finishes.
+			// TODO: A remedy would be to start a timer here, that would trigger a delayed ComputeCert event
+			//   (it's a one-liner here, but the corresponding delay parameter needs to be introduced in the params).
```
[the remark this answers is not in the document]
matejpavlovic: 1. In the previous implementation, the loop would also not continue, because the repetition was guarded by the `len(state.RequestStates) > 0` condition. Thus the loop would run once and then stop.
2. I think there is a problem with such an infinite loop though. While it doesn't bloat memory, it's busy-waiting and burns CPU at the maximum rate, since it _always_ generates something for a CPU to do. As a side effect, it also breaks the simulated time in tests, since it generates an infinite causal chain of events.

If we do put a timer here (but I think that even the current solution is ok for now), it can be a very very short one (even tens of milliseconds). This would give us basically the same advantage as a busy-wait loop, without killing the equivalent of a whole CPU core.

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
matejpavlovic: No, but we access them directly as `certificate.ReceivedSig`, not through a method of `certificate`. In fact, I don't know what the official convention is (should look that up), but I tend to expose fields accessed like this. But I'm ready to be corrected.

pkg/mempool/simplemempool/parts/computeids/computeids.go
matejpavlovic: Thanks, I'll remove it then.

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
matejpavlovic: As far as I understand, the situation where there are requests pending **and** certificates available is impossible. If it is possible, that's a bug, because it definitely shouldn't be - those requests should have been responded to with the existing certificates as soon as that situation occurred. Thus, if there is an empty certificate to be sent, it will certainly be the only one.

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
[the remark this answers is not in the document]
ranchalp: Fair enough! I reckon best would be some code at the mempool that specifically only returns once it has a batch, paired with a timer at the mempool to return an empty batch otherwise (perhaps with a flag set by the requester, i.e. the MSC, to have a small timer or a big one). Something of this sort, but for now we can leave it as is.

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
ranchalp: My understanding is that we should just keep them private if not used outside its package. But not a massive thing either way.

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
ranchalp: Good point, makes sense

pkg/availability/multisigcollector/parts/certcreation/certcreation.go
matejpavlovic: Fair enough. To be consistent, I un-exported everything that is not used outside of the package.

PR #417, opened 2023-04-23
Clean up producing and handling empty certificates

### LINKED ISSUE #360: Clean up empty decision/certificate

matejpavlovic:
Resolve #360.

We only want the availability module to both produce and consume availability certificates. Ts this is a much cleaner design than ISS producing availability certificates itself whenever agreement on a sequence number is aborted. We thus introduce the `empty` flag in the `DeliverCert` event emitted by ISS. It signifies that there is no certificate and the batch fetcher module (receiving the event) must not use the value stored in the `cert` field and simply deliver an empty transaction batch.

Builds on top of #416, which should be merged first.


pkg/iss/iss.go
```diff
@@ -687,27 +687,25 @@ func (iss *ISS) processCommitted() error {
...
 
-		var cert availabilitypb.Cert
-		if len(certData) > 0 {
-			if err := proto.Unmarshal(certData, &cert); err != nil {
+		var cert *apbtypes.Cert
+		if logEntry.Aborted {
+			iss.LeaderPolicy.Suspect(iss.epoch.Nr(), logEntry.Suspect)
+			cert = &apbtypes.Cert{Type: &apbtypes.Cert_Mscs{Mscs: &mscpbtypes.Certs{}}}
+			// TODO: Using a dummy MSC cert here. A nil value would be more natural, as this data is not to be used,
+			//  but currently nil values are not supported in generated types. Use nil when nil support is implemented.
```
ranchalp, line 699: If we have defined an empty cert as a slice with no values, would that not be better than having a nil value here?

ranchalp: Fair enough, we do need to support nil values either way for the code generator, but I think it is probably a better practice to semantically create a default empty cert than to use nil (that we already know can cause certain issues later on). Let's leave the comment there for now

PR #418, opened 2023-04-23
State update handlers in DSL module

### LINKED ISSUE #297: DSL UponCondition non-intuitive semantics
The DSL module has a convenient `UponCondition` method that can register handlers for arbitrary conditions. Those are evaluated each time a batch of events is applied to the module's state. This can lead to unintuitive behavior due to the batch-wise application of events, as the condition handler is not run after the application of every single event.

For example, a counter module maintaining an internal `counter` variable that is incremented on every event could register a condition handler for `counter == 10`. Due to (unpredictable) event application batching, the counter can go from 0 to 20 without having triggered the condition handler.

Such condition handling is still useful most of the time, but this non-intuitive behavior should be strongly pointed out in the documentation. Optionally, another type of condition handling could also be implemented that evaluates the condition after the application of every single event (with the obvious performance cost of many condition evaluations).

matejpavlovic:
Resolves #297.

### DIFF
--- a/pkg/bcb/bcbmodule.go
+++ b/pkg/bcb/bcbmodule.go
@@ -128,7 +128,7 @@ func NewModule(mc *ModuleConfig, params *ModuleParams, nodeID t.NodeID) modules.
 	})
 
 	// upon exists m != ⊥ such that #({p ∈ Π | echos[p] = m}) > (N+f)/2 and sentfinal = FALSE do
-	dsl.UponCondition(m, func() error {
+	dsl.UponStateUpdate(m, func() error {
 		if len(state.echoSigs) > (params.GetN()+params.GetF())/2 && !state.sentFinal {
 			state.sentFinal = true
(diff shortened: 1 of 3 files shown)


pkg/bcb/bcbmodule.go
```diff
@@ -128,7 +128,7 @@ func NewModule(mc *ModuleConfig, params *ModuleParams, nodeID t.NodeID) modules.
 	})
 
 	// upon exists m != ⊥ such that #({p ∈ Π | echos[p] = m}) > (N+f)/2 and sentfinal = FALSE do
-	dsl.UponCondition(m, func() error {
+	dsl.UponStateUpdate(m, func() error {
```
xosmig, line 131: Why not `UponStateUpdates`?

xosmig (approved):
LGTM

pkg/bcb/bcbmodule.go
matejpavlovic: Oh yes, forgot to update that. Thanks!

PR #419, opened 2023-04-24
codegen: Distinguish between `bytes` and repeated types

### LINKED ISSUE #356: Tagging `bytes` protobuffers with mir type results in wrong conversion

abread:
Also starts using them in a few places (batchdbpb, mempoolpb).

Closes #356


matejpavlovic (approved):
Thanks André, this is very useful!

It introduces some type conversions in the generated files that are not strictly necessary, so I disabled the linter for generated code.

PR #434, opened 2023-05-05
Orderers cleanup

ranchalp:
Refactor code (a bit more)


ranchalp:
Anticipating your potential comment, I moved as much as possible from proto messages to go, but I figured that if some proto message must remain (because of it being used as an actual event) then it is better to leave them as the context of `hasherpbdsl` and `cryptopbdsl` `UponResult` methods, than to create new structs with the same structure as the proto messages (less boilerplate).

pkg/orderers/pbftprotobufs.go
```diff
@@ -3,61 +3,18 @@ Copyright IBM Corp. All Rights Reserved.
...
-}
-
-func catchUpResponseHashOrigin(preprepare *pbftpb.Preprepare) *ordererpb.HashOrigin_Pbft {
-	return &ordererpb.HashOrigin_Pbft{
-		Pbft: &pbftpb.HashOrigin{Type: &pbftpb.HashOrigin_CatchUpResponse{CatchUpResponse: preprepare}},
-	}
-}
-
 // ============================================================
 // Serialization
```
matejpavlovic, line 19: Consider renaming this file, as it is only about serialization now, not about protobufs.

matejpavlovic (approved):
Amazing, the code is way cleaner and simpler, _and_ there is less of it!

pkg/orderers/pbftprotobufs.go
ranchalp: Done!

PR #435, opened 2023-05-05
Trantor modules cleanup

### LINKED ISSUE #431: Get rid of `iss.DefaultModules` and clean up creating a new Trantor instance
The `iss.DefaultModules` function used to make sense, but has since been progressively stripped off more and more modules. Currently it almost only checks whether some of the module dependencies are satisfied.

Either remove this function completely or, potentially, consider renaming it to something like `CheckModuleDependencies`, clean it up and introduce similar functions to to other modules, that simply check if all the modules in the module's specific `ModuleConfig` are indeed included in a set of modules.

Related with @ranchalp's comment in #428:

Why not have all default modules inside `modulesWithDefaults` directly? i.e.:
```
	// Factory module with instances of the checkpointing protocol.
    checkpointing := checkpoint.Factory(common.DefaultModuleConfig(), ownID, logging.Decorate(logger, "CHKP: "))


    // PBFT module with instances of the pbft protocol as segments to be called by ISS.
    ordering := orderers.Factory(orderers.DefaultModuleConfig(), params.Iss, ownID, hashImpl, cryptoImpl, …	
…
    batchFetcher := batchfetcher.NewModule(
   	…
    )
```

And why not have `hashImpl := crypto.SHA256` configurable? (but with this as part of a default like above?)

_Originally posted by @ranchalp in https://github.com/filecoin-project/mir/issues/428#issuecomment-1529281677_

matejpavlovic:
Cleans up the instantiation and configuration of modules within the Trantor SMR system.

Resolves #431.


pkg/availability/multisigcollector/multisigcollector.go
```diff
@@ -39,7 +28,7 @@ type ModuleParams = common.ModuleParams
 // Whenever an availability certificate is requested, it pulls a batch from the mempool module,
 // sends it to all replicas and collects params.F+1 signatures confirming that
 // other nodes have persistently stored the batch.
-func NewModule(mc *ModuleConfig, params *ModuleParams, logger logging.Logger) (modules.PassiveModule, error) {
+func NewModule(mc ModuleConfig, params *ModuleParams, logger logging.Logger) (modules.PassiveModule, error) {
```
ranchalp, line 31: Why deal directly with the struct and not the pointer? Any reason in particular? I suppose one reason is that the function should not change the struct, so better for it to deal with a copy. The drawback is performance: although the structs are not incredibly huge each call with `moduleConfig` copies the struct. 

Feel free to resolve with whatever decision upon reading this comment.

pkg/checkpoint/common/moduleconfig.go
```diff
@@ -1,25 +0,0 @@
-package common
```
ranchalp, line 1: I get why this would make sense in common, but as noted before I think there is a benefit to having the same skeleton structure in all modules (even if it is perhaps an overkill for this one), and other modules with more logic that benefit from splitting the logic into multiple `internal/parts/subpackages` (like mempool or availability) must in fact separate `moduleConfig` into a different subpackage than the main file of the module to avoid a cyclic import. 

Same as my other comment, feel free to resolve after reading this comment.

ranchalp (commented):
Other than the 2 comments, LGTM!

pkg/availability/multisigcollector/multisigcollector.go
matejpavlovic: The particular reason was that it slightly simplified the code (there actually were places where explicit copies were necessary) and I think it might even have been the case that while in most places it was passed by reference, it was occasionally still passed by value. I wanted to unify this. And I think that it makes more sense and is safer to pass it by value, so its contents are not overwritten by mistake somewhere. Regarding performance, as you said, the structs are very small, but, more importantly, for all practical performance purposes, they are used as good as never, so performance is definitely not a concern.

pkg/checkpoint/common/moduleconfig.go
matejpavlovic: I see the point. In that case, let's create a package named `config` or `params` or something like that and use it consistently. But I suggest putting this in the cleanup backlog, as this is a minor issue and a matter of polishing.

PR #446, opened 2023-05-12
codegen: Add support for nil arg to Pb(arg) and FromPb(arg)

### LINKED ISSUE #401: Proto-tagged maps and slices do not tolerate nil values
Struct fields of a proto message that are maps or slices and tagged (`mir.key_type` and `mir.value_type` for maps, or `mir.type` for the inner type of the slice) do not tolerate nil values.

A similar behaviour occurs the methods `...FromPb` (from the generated `types.mir.go`) on slices and certs whose content is empty.

ranchalp:
Closes #401


ranchalp:
Reopened, see #454

PR #448, opened 2023-05-12
Get rid of unused TxIDString and BatchIDString types

### LINKED ISSUE #430: Remove deprecated use of duplicated types due to now resolved issue #356
for example, there is no need anymore of using `t.BatchIdString` as we can properly tag the repeated bytes proto field `batch_id`. Not sure if there are more examples of this, that is also part of this issue to find out.

ranchalp:
Closes #430


matejpavlovic (approved):
Thanks Alex!
Feel free to apply or not apply the suggestions and merge.

PR #449, opened 2023-05-12
Use empty messages for testing

matejpavlovic:
We used arbitrary (but very concrete) messages in tests, because nil values were not supported in event and message serialization. Now that nil values are serialized and deserialized properly, we can use simpler messages in test.

@ranchalp when the nil value support is fixed, you can rebase this branch on top of it (then the tests should pass) and merge right away (if the tests do indeed pass).


ranchalp:
I did not have permission to push to a branch of your fork, so I rebased on a [branch of mine instead](https://github.com/filecoin-project/mir/pull/454).

matejpavlovic:
Subsumed by #454.

PR #459, opened 2023-05-22
Weighted voting

matejpavlovic:


pkg/trantor/testing/smr_property_test.go
```diff
@@ -120,6 +120,12 @@ func checkEventTraces(eventLogFiles map[t.NodeID]string, transactionsSubmitted i
 
 	// Check if each node delivered as many transactions as were submitted.
 	for _, db := range deliveredBatches {
+
+		// Skip this check if no transactions were submitted.
+		if transactionsSubmitted == 0 {
+			break
+		}
```
ranchalp, line 127: Curious about this, can you elaborate on how the changes on quorum size calculation causes an error here?

pkg/trantor/testing/smr_property_test.go
```diff
@@ -129,6 +135,9 @@ func checkEventTraces(eventLogFiles map[t.NodeID]string, transactionsSubmitted i
 		if cnt.Error() {
 			return es.Errorf("could not read number of delivered batches: %w", cnt.E)
 		}
+		if cnt.V == nil {
+			return es.Errorf("no transactions were delivered")
+		}
```
ranchalp, line 140: Curious about this, can you elaborate on how the changes on quorum size calculation causes an error here?

pkg/orderers/internal/common/pbftviewchangestate.go
```diff
@@ -53,7 +54,7 @@ func NewPbftViewChangeState(seqNrs []tt.SeqNr, membership []t.NodeID) *PbftViewC
 	}
 
 	return &PbftViewChangeState{
-		NumNodes:          len(membership),
+		Membersip:         membership,
```
ranchalp, line 57: Typo: Membership

pkg/iss/leaderselectionpolicy/blacklist.go
```diff
@@ -28,15 +28,16 @@ func NewBlackListLeaderPolicy(members []t.NodeID, MinLeaders int) *BlacklistLead
 	}
 }
 
-// Leaders always returns the whole membership for the SimpleLeaderPolicy. All nodes are always leaders.
+// Leaders returns a list of least recently suspected leaders
```
ranchalp, line 30: Typo: return

pkg/iss/leaderselectionpolicy/leaderselectionpolicy_test.go
```diff
@@ -143,9 +129,21 @@ func TestBlacklistLeaderPolicy_EmptySuspected(t *testing.T) {
...
 }
+
+func testMembership(nodeIDs []types.NodeID) *trantorpbtypes.Membership {
+	m := trantorpbtypes.Membership{Nodes: map[types.NodeID]*trantorpbtypes.NodeIdentity{}}
+	for _, nodeID := range nodeIDs {
+		m.Nodes[nodeID] = &trantorpbtypes.NodeIdentity{
+			Id:     nodeID,
+			Addr:   "",
+			Key:    nil,
+			Weight: 1,
```
ranchalp, line 145: As a note: having (not necessarily in the lsp package) tests for variable weight sizes would not be bad.

pkg/availability/multisigcollector/internal/parts/certverification/certverification.go
```diff
@@ -130,8 +132,12 @@ func verifyCertificateStructure(params *common.ModuleParams, cert *apbtypes.Cert
 	for _, mscCert := range mscCerts {
 
 		// Check that the certificate contains a sufficient number of signatures.
```
ranchalp, line 134: Update comment

pkg/availability/multisigcollector/internal/parts/certverification/certverification.go
```diff
@@ -148,7 +154,8 @@ func verifyCertificateStructure(params *common.ModuleParams, cert *apbtypes.Cert
 		}
 
 		// Check that signers are members.
-		if !sliceutil.ContainsAll(params.AllNodes, mscCert.Signers) {
+		// TODO: This check can be extremely inefficient (quadratic in membership size), especially at large scale.
+		if !sliceutil.ContainsAll(maputil.GetKeys(params.Membership.Nodes), mscCert.Signers) {
```
ranchalp, line 158: Good point, gladly we can just change the code of the containsAll function. We can make it linear using a completely reasonable extra amount of space with a map:
```
// ContainsAll returns true if the slice contains all elements of the given slice.
func ContainsAll[T comparable](ts []T, ts2 []T) bool {
	elements := make(map[T]struct{})
	for _, t := range ts {
		elements[t] = struct{}{}
	}

	for _, t := range ts2 {
		if _, ok := elements[t]; !ok {
			return false
		}
	}
	return true
}
```

(Can be made in this PR if you want or later on, but then it would be good to add an issue to the backlog at least)

pkg/trantor/testing/smr_property_test.go
matejpavlovic: I think it has to do with how rxgo works. Here I take an observable that contains all the delivered batches, map each batch to an int64 representing the number of included transactions, and then take the sum of all those numbers (I really like the concept of reactive programming):
```go
cnt, err := db.Map(func(_ context.Context, b interface{}) (interface{}, error) {
			return int64(len(b.(*batchfetcherpb.NewOrderedBatch).Txs)), nil
		}).SumInt64().Get()
```
This returns an objectIf there are only empty batches, this returns an object (`cnt`) and `cnt.V` is the value we're looking for (`cnt` holds additional information about potential errors). As long as there is even a single batch, this works (if it is empty, the value is 0). However, for some reason, if no batch has been delivered whatsoever, the above construct doesn't return a `cnt` object with the value 0, but `nil`. And then `cnt.V` segfaults.

This is a really small corner case and is not really relevant for tests that deliver any batches (even empty). But like this the error is handled properly at least, instead of the test segfaulting (since now we check whether `cnt == nil`). The check for `transactionsSubmitted == 0` is added since, technically, this particular test should pass even if no batches have been delivered at all.

pkg/trantor/testing/smr_property_test.go
matejpavlovic: See [comment above](https://github.com/filecoin-project/mir/pull/459#discussion_r1206432303). It doesn't have anything to do with the quorum size calculation, and it would have been cleaner to fix it in a different PR.

I just initially had a bug in the quorum computation that completely prevented the delivery of any batches, which made me discover the issue with the test.

pkg/iss/leaderselectionpolicy/blacklist.go
matejpavlovic: I see that it sounds a bit odd, but `Leaders` is a name of the function that returns a list.

pkg/orderers/internal/common/pbftviewchangestate.go
matejpavlovic: Fixed, thanks!

pkg/iss/leaderselectionpolicy/leaderselectionpolicy_test.go
matejpavlovic: Yes, once we define what weights are considered invalid, if any. I thought about that a bit, and realized that I don't see any fundamental reason to disallow any set of weights. The only exception is all weights in a membership being zero, which we probably do need to treat as a special case. I see 2 options here:
1. Declare such a membership invalid and complain / ignore it / crash / ...
2. Treat it as a special case and pretend that all weights are 1. I can imagine this being convenient for implementations that simply don't want to deal with any vote weights - it automatically falls back into counting nodes. However, it introduces additional complexity not just in code, but especially in the semantics.

pkg/availability/multisigcollector/internal/parts/certverification/certverification.go
matejpavlovic: Yes, let's make an issue of that.

pkg/availability/multisigcollector/internal/parts/certverification/certverification.go
matejpavlovic: Thanks, done.

PR #460, opened 2023-05-23
Sanitize messages and add handler for modules that do not tolerate transport events

### LINKED ISSUE #186: Sanitize external events and messages
Currently, the standard approach across Mir modules is to return an error upon receiving an unexpected event.
It can be exploited, for example, by sending a message to a module that does not expect to receive messages over the network (e.g., `"crypto"`).

ranchalp:
closes #186


ranchalp:
This took me more time that expected, as I started testing by sending a new tester message from the availability/0 module to Self in all other destinations, but no crash occurred. 

The reason was that the availability module treats events `UponMessageRecevied` (as it treats other subevents of `Message`) and such events already don't crash but return without error by the attempt when [casting them](https://github.com/filecoin-project/mir/blob/4b128afe6975b6429b58773f649cc8f1ecd43aad/pkg/pb/messagepb/dsl/messages.mir.go#L16-L19)

The fix in this PR is thus for all modules that may received a transport event without actually having any handler for any message. 

I had to fix unit tests that assumed modules should crash when sending messages they do not know about, but added a new test with a tester event that is not a message.

pkg/iss/iss.go
```diff
@@ -424,6 +424,11 @@ func New(
 		}
 		chkp := checkpoint.StableCheckpointFromPb(_chkp.Pb())
 
+		if err := chkp.SyntacticCheck(params); err != nil {
```
matejpavlovic, line 427: This is included in the verification of the checkpoint that is done later. I understand that we need that syntactic check first, since we access some fields before the full verification, but maybe we could still just run the full verification first (i.e., here). It is less efficient (especially if a malicious node makes us verify plenty of signatures uselessly - but it can do that anyway), but it's much cleaner to read and think about.

pkg/checkpoint/checkpoint.go
```diff
@@ -173,37 +174,38 @@ func (sc *StableCheckpoint) Verify(
 	params *issconfig.ModuleParams,
 	hashImpl crypto.HashImpl,
 	chkpVerifier Verifier,
+	membership *trantorpbtypes.Membership,
 	logger logging.Logger,
```
matejpavlovic, line 178: In fact, I would completely remove the logger from here (see later comments - added this one after them).

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
```diff
@@ -19,13 +19,27 @@ import (
...
-	_ *common.ModuleParams,
+	params *common.ModuleParams,
 	_ *common.State,
 ) {
 	mppbdsl.UponRequestTransactionIDs(m, func(txs []*trantorpbtypes.Transaction, origin *mppbtypes.RequestTransactionIDsOrigin) error {
-		txMsgs := make([]*hasherpbtypes.HashData, len(txs))
-		for i, tx := range txs {
-			txMsgs[i] = &hasherpbtypes.HashData{Data: serializeTXForHash(tx.Pb())}
+		if txs == nil || len(txs) > params.MaxTransactionsInBatch {
+			// Invalid request, ignore
```
matejpavlovic, line 27: This is not an external message (or could it be based on one?), but an internal event. If this check fails, that means that something is wrong with the implementation and we should return an error here.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
```diff
@@ -19,13 +19,27 @@ import (
...
+			return nil
+		}
+
+		txMsgs := make([]*hasherpbtypes.HashData, 0, len(txs))
+		for _, tx := range txs {
+			serializedTx := serializeTXForHash(tx.Pb())
+			if serializedTx == nil {
```
matejpavlovic, line 34: Also not sure if this can occur without an error in our code... (Not claiming it can't but I'm not sure.) In such a case, silently ignoring the transaction is very dangerous. Not to mention that it seems like a list of non-nil transactions would always have the same hash, regardless of how it is interleaved with nil transactions.
I think it would be better to return errors all through the mempool implementation whenever there is a nil transaction. We'd then only need to sanitize the input.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
```diff
@@ -19,13 +19,27 @@ import (
...
+				continue
+			}
+			txMsgs = append(txMsgs, &hasherpbtypes.HashData{Data: serializedTx})
+		}
+
+		// There should be at least one transaction, otherwise we would not have gotten here (see batchfetcher/UponDeliverCert)
```
matejpavlovic, line 40: Is this a sanity check? If yes, we should return an error if it fails.

pkg/orderers/internal/common/pbftsegmentchkp.go
```diff
@@ -108,6 +108,11 @@ func (chkp *PbftSegmentChkp) NodeDone(nodeID types.NodeID, doneDigests [][]byte,
 	// Store Done message
 	chkp.doneReceived[nodeID] = struct{}{}
 	strKey := chkp.aggregateToString(doneDigests)
+
+	if _, ok := chkp.doneMsgIndex[strKey]; !ok {
+		chkp.doneMsgIndex[strKey] = []types.NodeID{nodeID}
+	}
+
```
matejpavlovic, line 115: No need for this code. Appending to nil list works even without initialization.

pkg/orderers/internal/common/pbftslot.go
```diff
@@ -152,6 +152,10 @@ func (slot *PbftSlot) CheckCommitted() bool {
 }
 
 func (slot *PbftSlot) GetPreprepare(digest []byte) *pbftpbtypes.Preprepare {
+	if digest == nil || slot.Preprepare == nil {
```
matejpavlovic, line 155: The check for `slot.Preprepare == nil` is redundant.

pkg/orderers/internal/parts/goodcase/pbftgoodcase.go
```diff
@@ -277,6 +278,10 @@ func ApplyMsgPreprepare(
 	// Convenience variable
 	sn := preprepare.Sn
 
+	if preprepare.Data == nil {
+		return
```
matejpavlovic, line 282: Let's log a warning here.

pkg/orderers/internal/parts/goodcase/pbftgoodcase.go
```diff
@@ -334,6 +339,10 @@ func applyMsgPrepare(
 	logger logging.Logger,
 ) {
 
+	if prepare.Digest == nil {
+		return // Ignore Prepare messages with nil digest.
```
matejpavlovic, line 343: Let's log a warning here.

pkg/orderers/internal/parts/goodcase/pbftgoodcase.go
```diff
@@ -370,6 +379,10 @@ func applyMsgCommit(
 	logger logging.Logger,
 ) {
 
+	if commit.Digest == nil {
+		return
```
matejpavlovic, line 383: Let's log a warning here.

pkg/orderers/internal/parts/viewchange/pbftviewchange.go
```diff
@@ -627,6 +633,10 @@ func applyMsgNewView(
 		return
 	}
 
+	if len(newView.ViewChangeSenders) == 0 || len(newView.ViewChangeSenders) != len(newView.SignedViewChanges) {
+		return
```
matejpavlovic, line 637: Why are we checking for 0 separately here? The second part of the check intuitively seems enough, assuming that some other part of the code checks that there are sufficiently many SignedViewChanges.

Also, let's log a warning here.

matejpavlovic (changes requested):
Nice! Only small comments.
Thanks Alex!

pkg/iss/iss.go
ranchalp: Yes, you got my train of thought when I called it there. Let us do it as you say despite the slight potential inefficiency (We can always put the check of whether the node is part of the membership higher up with an extra check or so). Done.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: This is called as part of an external message received [UponProvideBatchMessageReceived](https://github.com/filecoin-project/mir/blob/8f352300a01bb71d1bc05cb62a92e0e105ad0862/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go#L128-L142)

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: My understanding is that returning an error in a dsl.Upon function halts execution, and we do not want that if a corrupted node sends us a corrupted batch with empty transactions.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: I think what you mean is then just halting this calculation and returning nil (because this can be triggered by an external event [UponProvideBatchMessageReceived](https://github.com/filecoin-project/mir/blob/8f352300a01bb71d1bc05cb62a92e0e105ad0862/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go#L128-L142).

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: Either way, just returning nil in the above conservation (i.e. as soon as a tx is nil) and checking with the check at the beginning that txs != nil would suffice. Done this.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: Done what I mentioned but feel free to comment.

pkg/orderers/internal/common/pbftsegmentchkp.go
ranchalp: Thanks! done

pkg/orderers/internal/common/pbftslot.go
ranchalp: Fair!

pkg/orderers/internal/parts/goodcase/pbftgoodcase.go
ranchalp: done!

pkg/orderers/internal/parts/goodcase/pbftgoodcase.go
ranchalp: Done.

pkg/orderers/internal/parts/goodcase/pbftgoodcase.go
ranchalp: Done!

pkg/orderers/internal/parts/viewchange/pbftviewchange.go
ranchalp: You are right, the code will take a wild journey until it eventually stops [here ](https://github.com/filecoin-project/mir/blob/8f352300a01bb71d1bc05cb62a92e0e105ad0862/pkg/orderers/internal/parts/viewchange/pbftviewchange.go#L212), but that suffices.

ranchalp: Done removed.

pkg/checkpoint/checkpoint.go
ranchalp: Fair, done!

pkg/dsl/dslmodule.go
```diff
@@ -230,9 +230,23 @@ func (m *dslModuleImpl) ApplyEvents(evs *events.EventList) (*events.EventList, e
...
 		return nil
 	}
 	return es.Errorf("unknown event type '%T'", ev.Type)
 }
+
+// The failExceptForInitAndTransport is just like failExceptForInit except
+// that if the module does not tolerate a transport event, it gracefully
+// ignores it. This prevents external nodes from crashing the node by sending
+// transport events to modules that do not tolerate it.
+func failExceptForInitAndTransport(ev *eventpb.Event) error {
```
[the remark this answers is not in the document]
ranchalp: Fair and done!

pkg/checkpoint/checkpoint.go
```diff
@@ -258,3 +266,58 @@ func hash(data [][]byte, hasher crypto.HashImpl) []byte {
...
+		return es.Errorf("invalid checkpoint: number of memberships %v does not match expected %v",
+			params.ConfigOffset+1,
+			len(sc.Memberships()))
+	}
+
+	if sc.StateSnapshot().EpochData.PreviousMembership == nil {
+		return es.Errorf("previous membership is nil")
+	}
+
+	if sc.Cert == nil {
```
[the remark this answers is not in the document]
ranchalp: Done.

pkg/checkpoint/checkpoint.go
```diff
@@ -229,6 +231,12 @@ func (sc *StableCheckpoint) verifyMembershipConsistency(logger logging.Logger) e
 			}
 		}
 	}
+
+	if sc.Epoch() > 0 && reflect.DeepEqual(sc.PreviousMembership(), membership) {
+		logger.Log(logging.LevelWarn, "Inconsistent sc.PreviousMembership() with membership")
```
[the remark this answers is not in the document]
ranchalp: Done I think by resolving previous comment.

pkg/checkpoint/checkpoint.go
```diff
@@ -173,37 +174,38 @@ func (sc *StableCheckpoint) Verify(
...
 	}
 
 	return nil
 }
 
 // verifyMembershipConsistency verifies that if the same node appears then the same parameters are always used
-func (sc *StableCheckpoint) verifyMembershipConsistency(logger logging.Logger) error {
+func (sc *StableCheckpoint) verifyMembershipConsistency(membership *trantorpbtypes.Membership, logger logging.Logger) error {
 	membershipConsistency := map[t.NodeID]*trantorpbtypes.NodeIdentity{}
 	for _, membership := range sc.Memberships() {
```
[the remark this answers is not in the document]
ranchalp: Fair, done.

pkg/checkpoint/checkpoint.go
```diff
@@ -173,37 +174,38 @@ func (sc *StableCheckpoint) Verify(
...
 		}
 	} else if len(sc.PreviousMembership().Nodes) > 0 {
-		logger.Log(logging.LevelWarn, "Ignoring starting checkpoint. Certificate not empty for first epoch.")
-		return es.Errorf("invalid starting checkpoint: certificate not empty for first epoch")
-	}
-
-	//verify that sufficient memberships are provided by the checkpoint
-	if len(sc.Memberships()) != params.ConfigOffset+1 {
-		return es.Errorf("invalid starting checkpoint: number of memberships does not match params.ConfigOffset")
+		logger.Log(logging.LevelWarn, "Ignoring checkpoint. Certificate not empty for first epoch.")
```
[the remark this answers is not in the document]
ranchalp: Done.

pkg/checkpoint/checkpoint.go
```diff
@@ -173,37 +174,38 @@ func (sc *StableCheckpoint) Verify(
...
+		logger.Log(logging.LevelWarn, "Ignoring starting checkpoint. Syntactic check failed.")
+		return es.Errorf("invalid starting checkpoint: %w", err)
+	}
+
 	// Only verify certificate if not the first epoch
 	if sc.Epoch() > 0 {
-		if err := sc.VerifyCert(hashImpl, chkpVerifier, sc.PreviousMembership()); err != nil {
-			logger.Log(logging.LevelWarn, "Ignoring starting checkpoint. Certificate not valid.",
+		if err := sc.VerifyCert(hashImpl, chkpVerifier, membership); err != nil {
+			logger.Log(logging.LevelWarn, "Ignoring checkpoint. Certificate not valid.",
```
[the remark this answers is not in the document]
ranchalp: Done.

pkg/checkpoint/checkpoint.go
```diff
@@ -173,37 +174,38 @@ func (sc *StableCheckpoint) Verify(
 	params *issconfig.ModuleParams,
 	hashImpl crypto.HashImpl,
 	chkpVerifier Verifier,
+	membership *trantorpbtypes.Membership,
 	logger logging.Logger,
 ) error {
 
+	if err := sc.SyntacticCheck(params); err != nil {
+		logger.Log(logging.LevelWarn, "Ignoring starting checkpoint. Syntactic check failed.")
```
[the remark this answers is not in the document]
ranchalp: That's fair. done.

ranchalp:
Thanks for the review @matejpavlovic ! just addressed all comments :slightly_smiling_face:

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
matejpavlovic: I see. But in that case, if we really do want to enforce this predicate on the received batch, I would prefer enforcing it at the reception of the message.

In principle, from the point of view of the ID computation code, IDs can be computed for both empty batches and large batches, so I find it much cleaner to not enforce this condition here. This part of the code does not seem to rely on it, so it should not enforce it.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
matejpavlovic: What I meant was to
1. Return an error here
2. Make sure that this event is never triggered with `nil`s in the transaction list. (Might be less efficient, but is cleaner.) Should that be about to happen because of an external message, catch it right there and ignore the message with a warning.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
matejpavlovic: Same point as above. If it can happen at [UponProvideBatchMessageReceived](https://github.com/filecoin-project/mir/blob/8f352300a01bb71d1bc05cb62a92e0e105ad0862/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go#L128-L142), catch it there. I think it's good practice to sanitize input as close to the input itself as possible, not deeper down the code path. Things getting silently ignored somewhere deep in the code is the best recipe for debugging nightmares.

pkg/orderers/internal/parts/viewchange/pbftviewchange.go
```diff
@@ -620,13 +626,19 @@ func applyMsgNewView(
...
+	logger logging.Logger,
 ) {
 
 	// Ignore message if the sender is not the primary of the view.
 	if from != state.Segment.PrimaryNode(newView.View) {
 		return
 	}
 
+	if len(newView.ViewChangeSenders) != len(newView.SignedViewChanges) {
+		logger.Log(logging.LevelDebug, "Ignoring NewView message with invalid view change data.")
```
matejpavlovic, line 638:
```suggestion
		logger.Log(logging.LevelWarn, "Ignoring NewView message with invalid view change data.")
```

matejpavlovic (changes requested):
Getting there!

ranchalp:
I cannot comment on [this review message](https://github.com/filecoin-project/mir/pull/460#discussion_r1212677269) and you are right on your review message (and I was right before I changed the code). I have readded the syntactic check to after the checkpoint is reconstructed from Pb, where it used to be before :).

pkg/iss/iss.go
ranchalp: I have readded the syntactic check to after the checkpoint is reconstructed from Pb, where it used to be before :). I also readded the simpler check of the node being part of the membership earlier again as an optimzation.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: MaxTransactionsInBatch is a parameter of the mempool that the availability module does not even know about (and should not have to imo). I think it is conceptually sensible to make the check in the mempool and not in the availability module. We can remove the check with nil, but at the moment it is also a condition that no nil transaction is given as a request. I am happy to remove the `== nil`. 

I have added a log in the event of that check returning nil.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: This relates to the previous comment. Sure I can make those checks in [UponProvideBatchMessageReceived](https://github.com/filecoin-project/mir/blob/8f352300a01bb71d1bc05cb62a92e0e105ad0862/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go#L128-L142), but the one for MaxTransactionsInBatch I am reluctant to move for as long as that parameter is part of the default config of the mempool only.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: Done for both [UponProvideBatchMessageReceived](https://github.com/filecoin-project/mir/blob/8f352300a01bb71d1bc05cb62a92e0e105ad0862/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go#L128-L142) and UponRequestSigMessageReceived (see commit for this). As mentioned, the upper bound check is for the moment still in mempool.

ranchalp:
Thanks for the comments! Went through them feel free to recheck!

ranchalp:
Not sure why lint is not working? it locally works.

pkg/iss/iss.go
matejpavlovic: Perfect. In fact, checking if the node is in the checkpoint's membership is not an optimization (I didn't realize you had removed that too). The checkpoint can be perfectly valid and still not include the receiving node in its membership, so this check cannot go inside the checkpoint's `Verify` method. But in ISS we still need to check it, because if we tried to restore state from a valid checkpoint where we are not a member, things would break.
You can resolve this convo after reading this, the code is good now as it is.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
```diff
@@ -19,13 +20,24 @@ import (
...
+	logger logging.Logger,
 	_ *common.State,
 ) {
 	mppbdsl.UponRequestTransactionIDs(m, func(txs []*trantorpbtypes.Transaction, origin *mppbtypes.RequestTransactionIDsOrigin) error {
-		txMsgs := make([]*hasherpbtypes.HashData, len(txs))
-		for i, tx := range txs {
-			txMsgs[i] = &hasherpbtypes.HashData{Data: serializeTXForHash(tx.Pb())}
+		if len(txs) > params.MaxTransactionsInBatch {
+			// Invalid request, ignore
+			logger.Log(logging.LevelWarn, "invalid request too big for mempool", "txs", txs)
```
matejpavlovic, line 30:
```suggestion
			logger.Log(logging.LevelWarn, "Ignoring invalid request: too big for mempool", "numTXs", len(txs))
```

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
matejpavlovic: That's fair. In that case, let's keep this and the `nil` check here. I'd still also print a warning though when ignoring the whole request because of a `nil` transaction (also suggested separately).

Also, feel free to resolve this one after reading.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
```diff
@@ -19,13 +20,24 @@ import (
...
+			return nil
+		}
+
+		txMsgs := make([]*hasherpbtypes.HashData, 0, len(txs))
+		for _, tx := range txs {
+			serializedTx := serializeTXForHash(tx.Pb())
+			if serializedTx == nil {
+				return nil
```
matejpavlovic, line 38:
```suggestion
				return nil
				logger.Log(logging.LevelWarn, "Ignoring invalid request: contains nil transaction", "offset", i)
```

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
```diff
@@ -19,13 +20,24 @@ import (
...
-		for i, tx := range txs {
-			txMsgs[i] = &hasherpbtypes.HashData{Data: serializeTXForHash(tx.Pb())}
+		if len(txs) > params.MaxTransactionsInBatch {
+			// Invalid request, ignore
+			logger.Log(logging.LevelWarn, "invalid request too big for mempool", "txs", txs)
+			return nil
+		}
+
+		txMsgs := make([]*hasherpbtypes.HashData, 0, len(txs))
+		for _, tx := range txs {
```
matejpavlovic, line 35:
```suggestion
		for i, tx := range txs {
```

pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go
```diff
@@ -132,9 +134,17 @@ func IncludeBatchReconstruction(
...
 			return nil
+		} else if _, ok = req.Txs[batchID]; !ok {
+			// Ignore a message with a batchID that was not requested.
+			return nil
+		}
+
+		if txs == nil || sliceutil.Contains(txs, nil) {
+			// Ignore an empty txs list or a txs list with a nil tx.
+			return nil
 		}
```
matejpavlovic, line 148: I would log a warning on each of these occasions, saying that we ignore the message and why.

pkg/availability/multisigcollector/internal/parts/certcreation/certcreation.go
```diff
@@ -186,6 +186,11 @@ func IncludeCreatingCertificates(
 			logger.Log(logging.LevelWarn, "sender %s is not a member.\n", from)
 			return nil
 		}
+		if txs == nil || sliceutil.Contains(txs, nil) {
```
matejpavlovic, line 189:
```suggestion
		if len(txs) == 0 || sliceutil.Contains(txs, nil) {
```
(works also for `nil` and catches a non-nil empty list)

pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go
```diff
@@ -132,9 +134,17 @@ func IncludeBatchReconstruction(
...
-		if _, ok := state.RequestState[reqID]; !ok {
+		if req, ok := state.RequestState[reqID]; !ok {
 			// Ignore a message with an invalid or outdated request id.
 			return nil
+		} else if _, ok = req.Txs[batchID]; !ok {
+			// Ignore a message with a batchID that was not requested.
+			return nil
+		}
+
+		if txs == nil || sliceutil.Contains(txs, nil) {
```
matejpavlovic, line 145:
```suggestion
		if len(txs) == 0 || sliceutil.Contains(txs, nil) {
```
(works also for nil, and catches a non-nil empty list)

matejpavlovic (approved):
Very nice!
Only trivial changes left.

matejpavlovic:
As for the linting, let's not worry about that for this PR and see if the same error keeps occurring.

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: Done locally (as per previous comment)

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
ranchalp: Done locally as per previous comment

pkg/availability/multisigcollector/internal/parts/certcreation/certcreation.go
ranchalp: We talked about moving nil check to computeids.go, so not committing this

pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go
ranchalp: You suggested to move the nil check to computeids.go, so not committing this

pkg/mempool/simplemempool/internal/parts/computeids/computeids.go
matejpavlovic: In fact, this condition is wrong. I guess you meant `||` instead of `&&`.
But anyway, I still think that it's better to remove the `len(txs) == 0` check. I see no strong reason for the mempool to not work for empty batches.
What I meant by the `nil` check is checking that no transactino in the list is `nil` (the code just below). This, as you convinced me, we should keep.

pkg/availability/multisigcollector/internal/parts/certcreation/certcreation.go
matejpavlovic: Sorry I wasn't clear enough about that. What I meant by the `nil` check is checking if any contained transactions are `nil` (second part of this condition), not if the whole list is `nil`. As mentioned in the other comment, I'd suggest doing  only the first part here (`len(txs)` == 0), since it is the availability module that rejects empty lists, and not the mempool. The second part can be removed indeed, since the mempool implements it as well.

pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go
matejpavlovic: Sorry I wasn't clear enough about that. What I meant by the `nil` check is checking if any contained transactions are `nil` (second part of this condition), not if the whole list is `nil`. As mentioned in the other comment, I'd suggest doing  only the first part here (`len(txs)` == 0), since it is the availability module that rejects empty lists, and not the mempool. The second part can be removed indeed, since the mempool implements it as well.

matejpavlovic (changes requested):
Almost there!

ranchalp:
Alright then, the minor changes are pushed.

matejpavlovic (approved):
🎉🎉🎉
Great job Alex, ready to merge!
(I pushed a tiny fix of a log message misprint and rebased on top of main.)
I leave the pleasure of clicking the "merge" button to you @ranchalp 🙂

PR #473, opened 2023-07-03
Parametrize availability layer sub-certificate limit

### LINKED ISSUE #468: Checkpoint typo did not trigger FAILed tests, why?
@matejpavlovic  and I have recently found [this bug](https://github.com/filecoin-project/mir/blob/55a940a4ab7c5d0bdd97d908748556b8776c8cd8/pkg/checkpoint/checkpoint.go#L221) in which we missed the `!` before the `reflect.DeepEqual`. While this is an easy-to-solve typo (coming in a PR soon), it is not expected that all tests PASSed given this error when verifying the checkpoint. 

We need to know why is that. We came up with the possibility that this is because the introduction of the async multisig collector means that in the first epoch all the transactions in all tests are delivered, so no checkpoint actually need to be verified in the tests. It could be this or something else, but we need to figure it out and resolve it so that the tests are comprehensive again.

matejpavlovic:
The limit on the number of sub-certificates is not hard-coded to 5 any more, but is a proper parameter of the Trantor system.

Some tests were designed before the code was even generalized to support more than one sub-certificate (the limit effectively being 1) and stopped working after the code supported 5 sub-certificates by default. Tests that should have failed were passing due to this.

This test makes this limit a proper parameter that can be higher by default, but still set to 1 in tests.

Resolves #468


ranchalp (approved):
LGTM! Thanks for the swift solving :)

PR #474, opened 2023-07-04
Event recorder sync writing

matejpavlovic:
The event recorder can now be instantiated with an option that makes the recorder write all events synchronously within the call to `Intercept`, rather than in the background.


ranchalp (approved):
LGTM!

PR #475, opened 2023-07-04
Abort faulty availability certificates

matejpavlovic:
This fixes an issue where an invalid availability certificate that has been delivered made it (with data truncated to an empty byte slice) all the way to batch reconstruction and made the system fail. The system did not recognize it as empty, because the aborted flag was false.


ranchalp (approved):
Good line :)

ranchalp (approved):

PR #480, opened 2023-07-05
Fix error in test

ranchalp:
The error was only in the deployment of the tests, because of the list of nodeIDs not being sorted when retrieved from the map of weights. This paired with the uneven weights meant that in the end we crashed a replica that took enough power to prevent delivery of any transaction. 

A simple word fixes the bug.


matejpavlovic (approved):
Good catch, thanks a lot!
Can you reduce the time for that particular test back to something reasonable? I think it's 40 seconds now, which is probably too much. You can directly merge afterwards, no need for further review.

ranchalp:
Upon further testing 30 seconds is not enough for the test, but 40 is. So let us leave it at that.

PR #483, opened 2023-07-07
Add timeouts and batch size to mempool

matejpavlovic:
The mempool itself can now wait for a transaction batch to fill. After a request for a new batch, the mempool waits until one of:
- it has a full batch with the configured number of transactions
- it has a full batch with the configured maximum total payload size
- a configurable timeout


ranchalp (approved):
LGTM!

PR #489, opened 2023-07-12
Log basic information about event processing in Mir Node

matejpavlovic:
This is more important than it might seem. When testing performance of a Mir-based system, the system is pushed to the limit, event buffers actually get full and external event processing often needs to be paused. Information on the buffer states and processed events brings a lot of insight.

### DIFF
--- a/eventbuffer.go
+++ b/eventbuffer.go
@@ -68,3 +68,12 @@ func (wi *eventBuffer) AddEvents(events *events.EventList) error {
 	}
 	return nil
 }
+
+// Stats returns a map containing the number of buffered events for each module.
+func (wi *eventBuffer) Stats() map[t.ModuleID]int {
+	stats := make(map[t.ModuleID]int)
+	for mID, evts := range wi.buffers {
+		stats[mID] = evts.Len()
+	}
+	return stats
+}
--- a/eventbuffer.go
+++ b/eventbuffer.go
@@ -32,10 +32,10 @@ func newEventBuffer(modules modules.Modules) eventBuffer {
 	return wi
 }
 
-// AddEvents adds events produced by modules to the eventBuffer buffer.
+// Add adds events produced by modules to the eventBuffer buffer.
 // According to their DestModule fields, the events are distributed to the appropriate internal sub-buffers.
-// When AddEvents returns a non-nil error, any subset of the events may have been added.
-func (wi *eventBuffer) AddEvents(events *events.EventList) error {
+// When Add returns a non-nil error, any subset of the events may have been added.
+func (wi *eventBuffer) Add(events *events.EventList) error {
 	// Note that this MUST be a pointer receiver.
 	// Otherwise, we'd increment a copy of the event counter rather than the counter itself.
 	iter := events.Iterator()
--- a/eventbuffer.go
+++ b/eventbuffer.go
@@ -35,7 +35,7 @@ func newEventBuffer(modules modules.Modules) eventBuffer {
 // Add adds events produced by modules to the eventBuffer buffer.
 // According to their DestModule fields, the events are distributed to the appropriate internal sub-buffers.
 // When Add returns a non-nil error, any subset of the events may have been added.
-func (wi *eventBuffer) Add(events *events.EventList) error {
+func (eb *eventBuffer) Add(events *events.EventList) error {
 	// Note that this MUST be a pointer receiver.
 	// Otherwise, we'd increment a copy of the event counter rather than the counter itself.
 	iter := events.Iterator()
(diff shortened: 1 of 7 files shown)


eventbuffer.go
```diff
@@ -68,3 +68,12 @@ func (wi *eventBuffer) AddEvents(events *events.EventList) error {
 	}
 	return nil
 }
+
+// Stats returns a map containing the number of buffered events for each module.
+func (wi *eventBuffer) Stats() map[t.ModuleID]int {
```
ranchalp, line 73: Why is the event buffer called `wi`?

nodestats.go
```diff
@@ -0,0 +1,106 @@
...
+	t "github.com/filecoin-project/mir/pkg/types"
+	"github.com/filecoin-project/mir/pkg/util/maputil"
+)
+
+// ==============================================================================================================
+// Event dispatching statistics
+// ==============================================================================================================
+
+// eventDispatchStats saves statistical information about the dispatching of events between modules,
+// such the numbers of events dispatched for each module.
```
ranchalp, line 17: Such as*

nodestats.go
```diff
@@ -0,0 +1,106 @@
...
+// newDispatchStats returns a new eventDispatchStats object with all counters set to 0
+// and the last dispatch time set to the current time.
+func newDispatchStats(moduleIDs []t.ModuleID) eventDispatchStats {
+	stats := eventDispatchStats{
+		dispatchCounts: make(map[t.ModuleID]int),
+		eventCounts:    make(map[t.ModuleID]int),
+		numDispatches:  0,
+		lastDispatch:   time.Now(),
+	}
+	for _, moduleID := range moduleIDs {
```
ranchalp, line 34: This is not necessary, accessing a key in a map that is not there will return the [zero value](https://go.dev/doc/effective_go#maps).

nodestats.go
```diff
@@ -0,0 +1,106 @@
...
+// For each module, it includes the number of event lists dispatched to that module (d),
+// the total number of events in those lists (e),
+// and the number of events still buffered at the input of that module (b).
+func (ds *eventDispatchStats) CombinedStats(bufferStats map[t.ModuleID]int) []interface{} {
+	logVals := make([]interface{}, 0, len(ds.eventCounts)+2)
+	totalEventsDispatched := 0
+	totalEventsBuffered := 0
+	for mID, cnt := range ds.dispatchCounts {
+		logVals = append(logVals,
+			fmt.Sprint(mID), fmt.Sprintf("d(%d)-e(%d)-b(%d)", cnt, ds.eventCounts[mID], bufferStats[mID]),
```
ranchalp, line 59: missing `\n`?

nodestats.go
ranchalp: Also the make already resets the map to a whole new instance (I saw it is used during `flushStats`.

pkg/deploytest/helpers.go
```diff
@@ -30,5 +30,5 @@ func NewLogger(parentLogger logging.Logger) logging.Logger {
 	if parentLogger != nil {
 		return logging.Synchronize(parentLogger)
 	}
-	return logging.Synchronize(logging.ConsoleDebugLogger)
+	return logging.ConsoleDebugLogger // No need to synchronize, the ConsoleDebugLogger is already synchronized.
```
ranchalp, line 33: Correct, but the synchronize function already checks this and exits quickly if that's the case. I do not see a scenario where the logger won't be synchronized already here though.

ranchalp (changes requested):
Minor things! 
Also, is renaming `workItems` to `pendingEvents` relevant for this PR?

eventbuffer.go
matejpavlovic: Good catch, thanks! Fixed.

nodestats.go
matejpavlovic: Fixed.

nodestats.go
matejpavlovic: Actually, I didn't have it there before and added it later deliberately.
The point is that when I'm printing the stats (by looping through the map), I want all modules to be printed, even if they have been given 0 events.
In the last commit, a changed the printing even further, so the output is always consistently sorted by module ID.

nodestats.go
matejpavlovic: No, this is deliberate too. This string will be used as an argument to the `Log` function and should not end with a new line.

pkg/deploytest/helpers.go
ranchalp: (feel free to resolve)

matejpavlovic: Yes that is right. And if we were using an arbitrary logger (that could be synchronized or not synchronized) I'd absolutely leave it there. But since this is a concrete logger of which we know is always synchronized, synchronizing it again is confusing and makes the code harder to understand.

ranchalp:
LGTM!

ranchalp (approved):

matejpavlovic:
Thanks for the quick review!

> Also, is renaming workItems to pendingEvents relevant for this PR?

Yes, you're totally right, I should have put that separately.

PR #491, opened 2023-07-12
Wait for stat monitor goroutine

matejpavlovic:
This is only to enforce a clean shutdown of a node.

### DIFF
--- a/nodestats.go
+++ b/nodestats.go
@@ -88,6 +88,8 @@ func (n *Node) monitorStats(interval time.Duration, wg *sync.WaitGroup) {
 		select {
 		case <-n.workErrNotifier.ExitC():
 			ticker.Stop()
+			n.Config.Logger.Log(logging.LevelInfo, "Event processing monitoring finished.")
+			n.flushStats()
 			return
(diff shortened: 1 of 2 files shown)


nodestats.go
```diff
@@ -75,15 +76,16 @@ func (ds *eventDispatchStats) CombinedStats(bufferStats map[t.ModuleID]int) []in
...
+func (n *Node) monitorStats(interval time.Duration, wg *sync.WaitGroup) {
+	defer wg.Done()
+
 	ticker := time.NewTicker(interval)
 
-Loop:
 	for {
 		select {
 		case <-n.workErrNotifier.ExitC():
 			ticker.Stop()
```
ranchalp, line 87: just wondering, could we not print the stats here too? Not awfully useful, but could perhaps help with debugging once the node fails.

ranchalp (approved):
One wonder,(see comment), but approved.

nodestats.go
matejpavlovic: Smart! Done.

PR #492, opened 2023-07-12
Generalize selecting any key/value from maps

matejpavlovic:
This is not used yet, but will be in the benchmarking code.


ranchalp (approved):
Easy, sorry for the delay again!

PR #493, opened 2023-07-14
Externalize Checkpoint validation to dedicated module

ranchalp:
Make module in turn tolerate different implementations of a checkpoint validity checker, with a first implementation `ConservativeCVC` that simply rejects checkpoints that are too far ahead.


pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
```diff
@@ -0,0 +1,64 @@
...
+		return es.Errorf("nodeID not in membership")
+	}
+
+	// Check how far the received stable checkpoint is ahead of the local node's state.
+	chkpMembershipOffset := int(sc.Epoch()) - 1 - int(epochNr)
+	if chkpMembershipOffset <= 0 {
+		// Ignore stable checkpoints that are not far enough
+		// ahead of the current state of the local node.
+		return es.Errorf("checkpoint not far ahead enough")
+	}
```
matejpavlovic, line 54: I'd probably leave this part in the ISS code, since the checkpoint is not necessarily invalid. ISS just doesn't need it.

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+type ConservativeCVC struct {
+	issParams    *issconfig.ModuleParams
+	ownID        t.NodeID
+	hashImpl     crypto.HashImpl
+	chkpVerifier checkpoint.Verifier
+}
+
+// NewConservativeCVC returns a new ConservativeCVC. This checkpoint validity checker simply rejects checkpoints whose signatures cannot be verified because
```
matejpavlovic, line 22: Try to keep the lines not longer than 120 characters.

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
```diff
@@ -0,0 +1,64 @@
...
+// the node does not know the membership of the relevant epoch yet.
```
matejpavlovic, line 23: This makes total sense, and I'd add another CVC that does the same thing, but allows new checkpoints with a huge warning, as we do currently.

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
```diff
@@ -0,0 +1,64 @@
...
+func NewConservativeCVC(issParams *issconfig.ModuleParams, ownID t.NodeID, hashImpl crypto.HashImpl, chkpVerifier checkpoint.Verifier) *ConservativeCVC {
```
matejpavlovic, line 24: We only ever use `configOffset` from the ISS params, and the CVC is intended not to be part of ISS. And, in fact, we don't even need the `configOffset` value, since it can be derived from `len(memberships)` passed to `Verify`. I thus propose not using the `issconfig.ModuleParams` at all in the CVC. (But watch out for a potential +1 or -1 in with `len(memberships)`, as it contains the current membership as well - thus one more than `configOffset`.)

pkg/iss/iss.go
```diff
@@ -70,9 +72,6 @@ type ISS struct {
 	// Implementation of the hash function to use by ISS for computing all hashes.
 	hashImpl crypto.HashImpl
 
-	// Verifier of received stable checkpoints. This is most likely going to be the crypto module used by the protocol.
```
matejpavlovic, line 73: This is actually very nice. That we don't need to pollute the ISS module with the checkpoint verifier.

(Resolve this directly, not change requested :-) )

pkg/iss/iss.go
```diff
@@ -419,57 +417,29 @@ func New(
...
-			iss.logger.Log(logging.LevelWarn, "Ignoring checkpoint. Not in membership.",
-				"sender", sender, "memberships", chkp.Memberships())
+	cvcpbdsl.UponCheckpointValidated(iss.m, func(err error, c *validateChechkpointContext) error {
+		if err != nil {
+			iss.logger.Log(logging.LevelWarn, "Ignoring checkpoint. Validation failed:", err)
 			return nil
 		}
 
-		// Check how far the received stable checkpoint is ahead of the local node's state.
+		chkp := checkpoint.StableCheckpointFromPb(c.checkpoint.Pb())
```
matejpavlovic, line 438: This was probably a sort of brain fart of mine when I wrote the original code. But I now I don't see a reason to convert the checkpoint to PB and back like this...

pkg/trantor/moduleconfig.go
```diff
@@ -17,6 +18,7 @@ type ModuleConfig struct {
 	BatchDB       t.ModuleID
 	BatchFetcher  t.ModuleID
 	Checkpointing t.ModuleID
+	CVC           t.ModuleID
```
matejpavlovic, line 21: Same as previous comment. `ChkpValidator`? (or something similar, but more descriptive than CVC?) (not a strong opinion)

pkg/trantor/system.go
```diff
@@ -170,6 +172,13 @@ func New(
 		logging.Decorate(logger, "CHKP: "),
 	)
 
+	trantorModules[moduleConfig.CVC] = cvc.NewModule(moduleConfig.ConfigureCheckpointValidityChecker(), cvc.NewConservativeCVC(
```
matejpavlovic, line 175: Let's use the "permissive" one by default (mentioned in another comment).

matejpavlovic (changes requested):
Thanks for the quick implementation!
Most comments are very small and all are easy to fix.

pkg/checkpoint/checkpointvaliditychecker/api.go
```diff
@@ -0,0 +1,16 @@
...
+	trantorpbtypes "github.com/filecoin-project/mir/pkg/pb/trantorpb/types"
+	tt "github.com/filecoin-project/mir/pkg/trantor/types"
+)
+
+// The CVC interface represents an implementation of the checkpoint validity checker primitives inside the MirModule module.
+// It is responsible for verifying the validity of checkpoints.
+type CVC interface {
+
+	// Verify verifies the validity of the checkpoint
+	// Returns nil on success (i.e., if the given signature is valid) and a non-nil error otherwise.
```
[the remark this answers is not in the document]
ranchalp: Change committed locally for simplicity.

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (cc *ConservativeCVC) Verify(chkp *checkpointpbtypes.StableCheckpoint, epochNr tt.EpochNr, memberships []*trantorpbtypes.Membership) error {
+	sc := checkpoint.StableCheckpointFromPb(chkp.Pb())
+
+	// Check syntactic validity of the checkpoint.
+	if err := sc.SyntacticCheck(cc.issParams); err != nil {
+		return err
+	}
+
+	// Ignore checkpoint if we are not part of its membership
```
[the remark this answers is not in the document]
ranchalp: Locally changed for simplicity

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
ranchalp: Fair enough

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
ranchalp: soz, done.

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
ranchalp: PermissiveCVC, done (coming submission)

pkg/iss/iss.go
ranchalp: There are three types: DSL-type StableCheckpoint, PB-type StableCheckpoint, and manual go-code with custom functions StableCheckpoint. Here, we are converting from DSL-type to Pb() and giving that to the go-code function that takes Pb as input to convert to go-code type.

pkg/trantor/moduleconfig.go
ranchalp: sure, will do!

ranchalp: done

protos/eventpb/eventpb.proto
```diff
@@ -40,20 +41,21 @@ message Event {
...
+    batchfetcherpb.Event batch_fetcher    = 15;
+    threshcryptopb.Event thresh_crypto    = 16;
+      checkpointpb.Event checkpoint       = 17;
+         factorypb.Event factory          = 18;
+             isspb.Event iss              = 19;
+         ordererpb.Event orderer          = 20;
+          cryptopb.Event crypto           = 21;
+             apppb.Event app              = 22;
+       transportpb.Event transport        = 23;
+    checkpointvaliditycheckerpb.Event cvc = 24;
```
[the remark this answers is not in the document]
ranchalp: Done all around :)

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
ranchalp: For the Permissive CVC I will of course return nil instead

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
ranchalp: yes the -1 is part of the calculation in the verify. I will directly pass configOffset to the CVC.

pkg/checkpoint/checkpointvaliditychecker/conservativecvc.go
matejpavlovic: I think it would make even more sense to not even invoke any validation whatsoever, since we just don't need to look at an old checkpoint at all, regardless of whether it's valid or not.
But you are right that it might stay in the validity checker as a sanity check.

ranchalp: Agreed, that is what I have in the upcoming commit (give me 2 mins max)

pkg/iss/iss.go
matejpavlovic: Aaaah yes I see, my bad.

pkg/checkpoint/chkpvalidator/permissivecvc.go
```diff
@@ -0,0 +1,75 @@
...
+	if _, ok := sc.Memberships()[0].Nodes[pcv.ownID]; !ok {
+		return es.Errorf("nodeID not in membership")
+	}
+
+	chkpMembership := sc.PreviousMembership() // TODO this is wrong and it is a vulnerability, come back to fix after discussion (issue #384)
+	chkpMembershipOffset := int(sc.Epoch()) - 1 - int(epochNr)
+
+	if chkpMembershipOffset > pcv.configOffset {
+		// cannot verify checkpoint signatures, too far ahead
+		// TODO here we should externalize verification/decision to dedicated module (issue #402)
```
matejpavlovic: This TODO seems to be addressed by this code now.

pkg/checkpoint/chkpvalidator/permissivecvc.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	// We consider a checkpoint invalid if we are not part of its membership
+	// (more precisely, membership of the epoch the checkpoint is at the start of).
+	// Correct nodes should never send such checkpoints, but faulty ones could.
+	if _, ok := sc.Memberships()[0].Nodes[pcv.ownID]; !ok {
+		return es.Errorf("nodeID not in membership")
+	}
+
+	chkpMembership := sc.PreviousMembership() // TODO this is wrong and it is a vulnerability, come back to fix after discussion (issue #384)
```
matejpavlovic: I wouldn't call this a TODO any more, as it is supposed to stay like this.
It still deserves a comment though, let's just get rid of the TODO.

pkg/checkpoint/chkpvalidator/permissivecvc.go
```diff
@@ -0,0 +1,75 @@
...
+			"configOffset", pcv.configOffset,
+		)
+	} else {
+		chkpMembership = memberships[chkpMembershipOffset]
+	}
+
+	if chkpMembershipOffset > pcv.configOffset {
+		// cannot verify checkpoint signatures, too far ahead
+		return es.Errorf("checkpoint too far ahead")
+	}
```
matejpavlovic: Shouldn't the permissive checker allow this?

pkg/trantor/system.go
matejpavlovic: Seems like the `ConservativeCV` is still there.

matejpavlovic (commented):
Getting there!

pkg/checkpoint/chkpvalidator/permissivecvc.go
ranchalp: very good point :sweat_smile:

pkg/checkpoint/chkpvalidator/permissivecvc.go
ranchalp: Also very good point

pkg/checkpoint/chkpvalidator/permissivecvc.go
ranchalp: true, thanks for the catch and sorry for not realising

matejpavlovic (approved):
Nice work, ready to merge :-)

PR #497, opened 2023-07-22
Indexed list

matejpavlovic:
Data structure needed by the new mempool implementation (in a separate PR) so that it does not drop transactions at epoch boundaries.

### DIFF
--- a/pkg/util/indexedlist/indexedlist.go
+++ b/pkg/util/indexedlist/indexedlist.go
@@ -101,7 +101,7 @@ func (il *IndexedList[K, V]) appendElement(e *element[K, V]) {
 
 // prependElement prepends a given element to the start of the list if not already present (anywhere in the list).
 // ATTENTION: The list must be locked when calling this function!
-func (il *IndexedList[K, V]) prependElement(e *element[K, V]) {
+func (il *IndexedList[K, V]) prependElement(e *element[K, V]) { //nolint:unused
 
 	// Ignore element if one with the same key already exists.
 	if _, ok := il.index[e.key]; ok {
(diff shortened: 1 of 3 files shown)


pkg/util/indexedlist/indexedlist.go
```diff
@@ -0,0 +1,210 @@
...
+
+type element[K comparable, V any] struct {
+	prev *element[K, V]
+	next *element[K, V]
+	key  K
+	val  V
+}
+
+// IndexedList is a doubly-linked list with a map index.
+// Appending, prepending, and removal of arbitrary elements is in constant time and thread-safe.
```
ranchalp, line 18: Appending, preprending and removing ~of~ ?

pkg/util/indexedlist/indexedlist.go
```diff
@@ -101,7 +101,7 @@ func (il *IndexedList[K, V]) appendElement(e *element[K, V]) {
 
 // prependElement prepends a given element to the start of the list if not already present (anywhere in the list).
 // ATTENTION: The list must be locked when calling this function!
-func (il *IndexedList[K, V]) prependElement(e *element[K, V]) {
+func (il *IndexedList[K, V]) prependElement(e *element[K, V]) { //nolint:unused
```
ranchalp, line 139: Just wondering, why does lint complain here?

ranchalp (approved):
LGTM!

pkg/util/indexedlist/indexedlist.go
matejpavlovic: It was `removal of arbitrary elements`, but I changed that sentence now to sound a bit better.

pkg/util/indexedlist/indexedlist.go
matejpavlovic: `prependElement` is not used anywhere and is not exported, that's why it complains.
I actually adapted this data structure from the "bucket" implementation in old ISS, where prepending was necessary. But since it was already implemented, it cost me nothing to keep it here too, and it makes the data structure more complete.

PR #498, opened 2023-07-22
Make the Trantor mempool not drop transactions at epoch boundaries

matejpavlovic:
This PR builds on top of #497, which should be merged first, then this one rebased, and only then reviewed and merged.


matejpavlovic:
This PR has been replaced by #501 with a rebased version of this code.

PR #500, opened 2023-07-26
Garbage-collect batch DB

matejpavlovic:
The Batch DB also needs to be garbage-collected.
The current implementation stores all batches ever delivered in memory and thus runs out of memory rather quickly under high load.


matejpavlovic:
Replaced by #502 containing a rebased version of this PR.

PR #506, opened 2023-08-01
Use big.Int encoded as string for vote weights

matejpavlovic:
This fixes the issue with huge weights overflowing and causing wrong computation of quorums.


dnkolegov (commented):
It would be good to add a test that the code works if validators have very big weights.

matejpavlovic:
> It would be good to add a test that the code works if validators have very big weights.

Done.

pkg/trantor/testing/smr_test.go
```diff
@@ -188,8 +187,8 @@ func testIntegrationWithISS(tt *testing.T) {
 		13: {"Submit 100 fake transactions with 4 nodes in simulation, two of them crashed (not messages sent, yes received) and holding the supermajority of stake",
 			&TestConfig{
 				NodeIDsWeight: deploytest.NewNodeIDsWeights(4, func(id t.NodeID) types.VoteWeight {
-					idfloat, _ := strconv.ParseFloat(id.Pb(), 64)
-					return types.VoteWeight(math.Pow(2, idfloat)) // ensures last 2 nodes weight is greater than twice the sum of the others'
+					numericID, _ := strconv.ParseInt(id.Pb(), 10, 64)
```
ranchalp, line 190: not sure this is strictly needed (along pow2 function). AFAIS this is still uint64 (and fits in uint64), and using float would allow us to use the native math.pow. If we had this in BigInts then it would make more sense for me. But minor thing, if you have a good reason for this change, let us leave it.

pkg/util/membutil/membutil.go
```diff
@@ -11,11 +13,11 @@ import (
...
 
 func WeightOf(membership *trantorpbtypes.Membership, nodeIDs []t.NodeID) tt.VoteWeight {
-	w := tt.VoteWeight(0)
+	sum := big.NewInt(0)
 	for _, node := range maputil.GetValuesOf(membership.Nodes, nodeIDs) {
-		w += node.Weight
+		sum.Add(sum, node.Weight.BigInt())
 	}
-	return w
+	return tt.VoteWeight(sum.String())
```
ranchalp, line 20: Minor thing, but using sum.Text(16) (or the best would be [sum.Text(62)](https://pkg.go.dev/math/big#Int.Text), we can compact to the max the bigInt representation in string to reduce the size of a stable checkpoint. Probably not something to worry so much, but again, worth noting a "6n" optimization in a membership slice (which is sent over the network).

ranchalp (commented):
Minor comments only!

pkg/util/membutil/membutil.go
ranchalp: this is because 0-9 (and a-z A-Z) use each 8 bits, so we optimize if using more characters.

pkg/trantor/testing/smr_test.go
matejpavlovic: The main reason I wrote the custom `pow2` function for ints is that i've always been weary of converting floats to ints, even if they "should" be integers in the mathematical sense. Otherwise I'd have stuck with floats and just appended the string zeroes as I do now anyway.

pkg/trantor/types/types.go
```diff
@@ -9,10 +10,41 @@ import (
...
+	if _, ok := new(big.Int).SetString(vw.String(), 10); !ok {
+		return false
+	}
+	return true
+}
+
+// BigInt converts a VoteWeight (normally represented as a string) to a big.Int.
+// BigInt panics if the underlying string is not a valid decimal representation of an integer.
+// Thus, BigInt must not be called on received input without having validated VoteWeight by calling the IsValid method.
+func (vw VoteWeight) BigInt() *big.Int {
```
adlrocha, line 42: What about having a `BigInt()` that returns an error and doesn't panic, and then a `MustParseBigInt()` that panics? I am usually wary of assuming that a user (like me) will read the comments and use the function properly. I would assume that it can't error if I don't see an error in the signature.

(Weak opinion weakly held, feel free to disregard)

adlrocha (approved):
LGTM. Minor nit-pick, but looks good.

PR #510, opened 2023-08-24
crypto: Generate keys deterministically

abread:
crypto/ecdsa.GenerateKey is no longer deterministic as of Go 1.20: https://github.com/golang/go/issues/58637
and that breaks cryptography for `cmd/bench`

Switched to a different module, that supports deterministic key generation explicitly.
This module was [referred to in the issue](https://github.com/golang/go/issues/58637#issuecomment-1600627963) by its creator, who is a maintainer of Go cryptography.


matejpavlovic (approved):
Thanks Andre!

PR #516, opened 2023-09-19
README: List supported and WIP protocols

abread:
I'm not aware of any others apart from the existing ISS(-PBFT), and the WIP Alea-BFT and Granite.

Each listed protocol is linked to its respective implementation.


matejpavlovic (approved):
In the code, the package name ISS is still used, but that's only for historical reasons and needs to change eventually. The system implemented is actually Trantor (which evolved from an implementation of ISS).

I fixed that and updated the README page for Trantor too.

PR #517, opened 2023-09-19
orderers/pbft: rely on mempool to handle proposal timeouts

abread:
Currently ISS(PBFT) waits for a timer before requesting transactions to propose, but this means inducing an artificial latency on any proposal (1s by default).

This PR changes the PBFT orderer to immediately request transactions from the mempool, and sets the timeout in the mempool instead (the logic was already implemented, this was probably forgotten along the way).

Note: I did not investigate thoroughly whether the ProposalTimeout event in PBFT is still necessary, nor an alternative name for it that better fits its usage.

### DIFF
--- a/pkg/mempool/simplemempool/simplemempool.go
+++ b/pkg/mempool/simplemempool/simplemempool.go
@@ -27,7 +27,7 @@ func DefaultModuleParams() *ModuleParams {
 	return &ModuleParams{
 		MaxTransactionsInBatch: 1024,
 		MaxPayloadInBatch:      1024 * 1024, // 1 MiB
-		BatchTimeout:           100 * time.Millisecond,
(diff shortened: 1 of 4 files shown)


matejpavlovic:
Thanks for pointing this out, I see you really have a very good understanding of how the system works!
I was actually thinking about removing this timeout altogether, as you propose, but I decided to keep it.
Instead, I just set the timeout value to zero in this particular setup.

In case Trantor is used with a mempool that does not support timeouts (as is, in fact, the case when used with the Lotus Filecoin node software), this still might come in handy.

abread:
I see. Would it make sense to instead change the default timeout to 0 instead?

matejpavlovic:
Yes it absolutely would. Just created an issue for that. (#519)

PR #523, opened 2023-11-25
modified existing functions to return an EventList for the modified interceptor for visualization

Cath3876:
Different functions now return an EventList or an EventRecord such that the Events intercepted by the intercepter can also be modified by it.

may want to check my modifications of the implementations of the following to check nothing is broken unintentionally, as I might have overlooked something:
- type Interceptor in pkg/eventlog/interceptor.go 
- and type EventWriter in pkg/eventlog/eventwriter.go

### DIFF
--- a/pkg/eventlog/multiinterceptor.go
+++ b/pkg/eventlog/multiinterceptor.go
@@ -8,6 +8,10 @@ type repeater struct {
 	interceptors []Interceptor
 }
 
+func (r *repeater) InterceptWithReturn(events *events.EventList) (*events.EventList, error) {
+	panic("implement me")
+}
+
 func (r *repeater) Intercept(events *events.EventList) error {
 
 	// Avoid nil dereference if Intercept is called on a nil *Recorder and simply do nothing.
--- a/pkg/eventlog/multiinterceptor.go
+++ b/pkg/eventlog/multiinterceptor.go
@@ -8,26 +8,23 @@ type repeater struct {
 	interceptors []Interceptor
 }
 
-func (r *repeater) InterceptWithReturn(events *events.EventList) (*events.EventList, error) {
-	panic("implement me")
-}
-
-func (r *repeater) Intercept(events *events.EventList) error {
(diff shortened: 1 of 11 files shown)


matejpavlovic (commented):
Thank you and congrats to the first Mir pull request! :-)
Looks good in general, I left a few suggestions.

pkg/eventlog/multiinterceptor.go
```diff
@@ -8,26 +8,23 @@ type repeater struct {
...
-		return nil
+		return events, nil
 	}
 
 	for _, i := range r.interceptors {
-		if err := i.Intercept(events); err != nil {
-			return err
+		_, err := i.Intercept(events)
+		if err != nil {
+			return events, err
```
matejpavlovic, line 24: By convention, we usually return the zero value (`nil` in this case) when an error occurs. I think here `nil` would be better too.

pkg/eventlog/multiinterceptor.go
```diff
@@ -8,26 +8,23 @@ type repeater struct {
...
 	// For more explanation, see https://mangatmodi.medium.com/go-check-nil-interface-the-right-way-d142776edef1
 	if r == nil {
-		return nil
+		return events, nil
 	}
 
 	for _, i := range r.interceptors {
-		if err := i.Intercept(events); err != nil {
-			return err
+		_, err := i.Intercept(events)
```
matejpavlovic, line 22: In fact, contrary to what I wrote in an older comment, there is a totally meaningful way of defining the event modification semantics of the MultiInterceptor. We can simply feed the output of one interceptor to the input of the next one. This would correspond to a sort of "serial" connection of the interceptors and it would be the users responsibility to use a meaningful list of interceptors.
And the only code that needs to change to achieve that is this.

```suggestion
		events, err := i.Intercept(events)
```

matejpavlovic (changes requested):
Nice. Almost there!

matejpavlovic (approved):
Nice job, congrats to your first merged PR! :-)

PR #526, opened 2023-12-04
Added functions related to the debugger into pingpong

Cath3876:
I added the functions related to the new interceptor with the basic functionality to accept or decline events. In main.go of the pingpong application, I added a few lines to use this interceptor


matejpavlovic:
Thanks for the PR Catherine!

Before I go through it, could you please rebase the `debugger` branch on top of `visualization` and make only the actual changes appear in a single commit? It seems like all but one of the commits here have already been merged as part of #524. The usual workflow is the following:
- Pull changes from the upstream branch (`filecoin-project:visualization` in our case) into `andreaj00:visualization`
- Commit your code to a new branch from there (i.e., the `debugger` branch would extend the `visualization` branch) or rebase the branch with the code on top of `visualization`
- Only then push.

This way, the PR will only show the relevant changes (and accidental overwriting of code with some old version is avoided too).

Also, the linter is complaining about this commit. You can see the errors either by running `make lint` in the mir directory (you need to have `golangci-lint` installed for that) or directly on GitHub, in the "Files changed" tab of this PR.

I'll have a look at the actual code as soon as these little issues are fixed.

PR #528, opened 2023-12-07
Added function related to the debugger

Cath3876:
I added the functions related to the new interceptor with the basic functionality to accept or decline events. In main.go of the pingpongapplication, I added a few lines to use this interceptor.


pkg/debugger/debugger.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// writerFactory creates and returns a WebSocket-based event writer
+// It determines the port for the WebSocket server based on the given nodeID and a base port value
+func writerFactory(_ string, nodeID t.NodeID, _ logging.Logger) (eventlog.EventWriter, error) {
+	ownPort, err := strconv.Atoi(string(nodeID))
+	if err != nil {
+		panic(err)
+	}
+	ownPort += BasePort
```
matejpavlovic, line 209: It would be a bit cleaner to ignore the node ID here and pass the port number to the writer in a different way. For example by defining this function in-line when creating the interceptor, and using a closure to pass the port number directly.

samples/pingpong/main.go
```diff
@@ -27,6 +28,12 @@ func main() {
 
 	// Get own ID from command line.
 	ownID := t.NodeID(os.Args[1])
+	// Get 0 for no debugger, 1 for debugger
```
matejpavlovic, line 31: an optional `-d` as the first argument would be more intuitive to use. I.e.:
if the first argument is `-d`, use debug mode and look for the ID in the second argument. Otherwise, treat the first argument as the ID.
Alternatively, you could also use `cobra` (see other applications in the mir repo), but that still seems to be an overkill for this pingpong demo.

matejpavlovic (changes requested):
Thanks for the PR! I made some comments, pls. have a look. I was sometimes rather brief, so if anything is not clear, please let me know immediately and I'll explain more.

pkg/debugger/debugger.go
```diff
@@ -0,0 +1,211 @@
...
+			if err != nil {
+				panic(err)
+			}
+
+			fmt.Println("WebSocket connection established")
+
+			// Update the attribute of the WSWriter object with the established connection
+			wsWriter.conn = conn
+			// go routine for incoming messages
+			go func() {
```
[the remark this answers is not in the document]
Cath3876: If I understand correctly, 2 levels of nested goroutines would enable nodes to have multiple websocket connections, which we do not need, so I changed it.

matejpavlovic (approved):
Thank you for the fixes! There are only two minor things left, very easy to address, we can merge when you have them fixed.

pkg/debugger/debugger.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil
+	}
+	return wsw.conn.Close()
+}
+
+// Write sends every event to the frontend and then wits for a message detailing how to proceed with that event
+// The returned EventList contains the accepted events
+func (wsw *WSWriter) Write(list *events.EventList, _ int64) (*events.EventList, error) {
+	for wsw.conn == nil {
+		fmt.Println("No connection.")
```
matejpavlovic, line 78: Use a logger instead here. A pointer to a logger can be saved in a field of the `WSWriter`, set at creation, and accessed here through `wsw.logger`, for example.

matejpavlovic:
I pushed another commit, simplifying a part of the code (avoiding conversions from `interface{}` to `string`). It was simpler to just do then write about it, so here it is.

PR #527, opened 2023-12-05
Create frontend to connect to MIR nodes to accept or decline new logs

andreaj00:
Create frontend to connect to MIR nodes to accept or decline new logs. See attached images for examples.

The MIR nodes should run on port 8080+<membership number>

Use cases:
1) Connecting to non-existing ws port
image

2) Connecting to existing ws port and starting to receive log messages
image

3) Clicking on 'Decline' -> decline the incoming log message and receive the new log message
image

4) Clicking on 'Accept' -> accept the incoming log message and store it in "Accepted Logs" and receive the new log message
image

PR #533, opened 2023-12-30
Frontend ~ editable text

andreaj00:
**NOTE, Merge first the pull request: https://github.com/filecoin-project/mir/pull/527**

### **New Feature:**
Added the possibility to modify the logs that each MIR node sends by modifying the incoming message. (For now, only the JSON leaves are modifiable. )
All the JSON types have been taken care of:
- it's important to notice that once a 'null' field is clicked to modify, it will become a 'string type afterward'. 
- For other types, the type will be preserved if the newly modified value is acceptable for that specific type. For example, inserting 'true' into a boolean field will remain a boolean, but using 'tru' will result in it becoming a string field.

### **Usage Example:**
Modifiable incoming log (highlighted by the border line):
image

Once clicked it becomes editable as a string:
image

By clicking '**Replace**' the modified version will be saved (and we will receive the next log in the 'Incoming Log' section):
image

If we click simply '**Accept**' we will accept the **original** version of the incoming log  (and we will receive the next log in the 'Incoming Log' section):
image
image
image

If we click '**Decline**' will just decline and drop the message (even if modified) (and we will receive the next log in the 'Incoming Log' section)

### DIFF
--- a/frontend/index.css
+++ b/frontend/index.css
@@ -81,3 +81,15 @@ button {
     margin-top: 5px;
     overflow: auto;
 }
+
+/* EditableText*/
+/* Style for spans with text content */
+.editable-text-span {
+    border: 1px solid #000;
+}
+
+/* Style for empty spans with padding */
+.editable-text-span:empty {
+    padding-left: 10px;
+    padding-right: 10px;
+}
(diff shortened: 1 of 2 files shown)


matejpavlovic:
Finally almost ready to merge this @andreaj00 ! :-)
It just needs to be rebased on top of the `visualization` branch. There are some conflicts that can probably be trivially resolved (by you, not by me, even though I tried to have a look at it myself).
I'll merge it as soon as the conflicts are resolved.

PR #535, opened 2023-12-30
Frontend ~ sync switch

andreaj00:
**NOTE, Merge first the pull request: https://github.com/filecoin-project/mir/pull/527**

### **New Feature:**
Add the possibility to automatically accept the incoming logs from the MIR node when 'Sync' option selected:

### **Usage Example:**
By default, we are using the 'Sync' mode where we are waiting for the user input (Accept/Decline):
image
When changing the toggle, we will go to 'Async' mode where the incoming logs will be accepted by default:
image

### DIFF
--- a/frontend/index.css
+++ b/frontend/index.css
@@ -106,3 +106,81 @@ button {
     padding-left: 10px;
     padding-right: 10px;
 }
+
+/* Sync/Async connection switch */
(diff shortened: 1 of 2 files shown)


matejpavlovic (approved):
LGTM!

PR #541, opened 2024-02-09
replace functionality - backend

Cath3876:
I added the necessary code to be able to replace an event with one that is sent from the frontend.
protojson.unmarshal only works if the  value types in the new event all match the ones from original event, else an error is returned.
I had to use protojson for both the mashalling and unmarshalling, for it to work correctly. Thus, the format of the message sent to the frontend has changed, messing with the display of the event in the HTML file. Instead of displaying a structured tree where only the leaves can be modified, it now appears as a single string that can be modified in its entirety.

### DIFF
--- a/cmd/bench/cmd/client.go
+++ b/cmd/bench/cmd/client.go
@@ -38,7 +38,7 @@ var (
 	clientCmd = &cobra.Command{
 		Use:   "client",
 		Short: "Generate and submit transactions to a Mir cluster",
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: func(_ *cobra.Command, _ []string) error {
 			return runClient()
 		},
 	}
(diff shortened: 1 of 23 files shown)


matejpavlovic:
I still think that using protojson is the easiest option forward, even if it requires adapting the frontend. I expect the necessary adaptations to be very small.

> Instead of displaying a structured tree where only the leaves can be modified, it now appears as a single string that can be modified in its entirety.

@andreaj00 could you please see how the frontend needs to be modified to support protojson-generated inputs? Thanks!

matejpavlovic:
I still think that using protojson is the easiest option forward, even if it requires adapting the frontend. I expect the necessary adaptations to be very small.

> Instead of displaying a structured tree where only the leaves can be modified, it now appears as a single string that can be modified in its entirety.

@andreaj00 could you please see how the frontend needs to be modified to support protojson-generated inputs? Thanks!

matejpavlovic (approved):
Looks good, I only had one small issue with the code (the given timestamp was ignored and a new one re-generated), but I fixed it myself.

Let's wait for @andreaj00 to confirm that the frontend can deal with protojson output before merging.

PR #543, opened 2024-02-21
Frontend ~ Stacked log

andreaj00:
Converted the Accepted Logs list into a stack with the most recent message on top and added a border around each accepted message

image


matejpavlovic (approved):
LGTM, thanks!

PR #544, opened 2024-02-21
Debugger ~ No connection message

andreaj00:
Changed the waiting connection message of the debugger from "No Connection" to "Waiting for interface connection to proceed"

image


matejpavlovic (commented):
👍

